Issue #5, opened 2020-04-09
tags not rendered in Message Object


naderafshari:
Tags are rendered in Operations Objects (at the bottom of the section as dark boxes) but they do not seem to be rendered for Message objects. Attached is a screenshot of using tags in a message.

![asyncAPIplayground1](https://user-images.githubusercontent.com/18668042/78948488-5a238780-7a7d-11ea-9d53-79b35fb22955.png)

derberg:
@naderafshari how do you see it renders, I quickly mocked something like
<img width="505" alt="Screenshot 2020-04-12 at 19 14 13" src="https://user-images.githubusercontent.com/6995927/79075235-e0e09a80-7cf1-11ea-9af9-6650961c9360.png">

on hover over `test1` you would see a tooltip with description. What do you think?

naderafshari:
Looks great to me thank you @derberg

derberg:
@naderafshari please have a look and test this PR https://github.com/asyncapi/html-template/pull/8

derberg:
Fixed by https://github.com/asyncapi/html-template/pull/8

PR #8, opened 2020-04-15
feat: add support for tags  externalDocs url

### LINKED ISSUE #6: externalDocs are not rendered
It appears that externalDocs is not rendered in any of the Objects - Operations, Message, Tag etc.
The externalDocs object has a description and a url field (no name field). One way to implement this would be to make the description into a hyperlink pointing to the url. Another way would be to render a button showing the name of the parent object (for example, the name of the parent tag - after tag issue https://github.com/asyncapi/html-template/issues/5 is fixed) in the button and pointing to the url and the description showing as text under the button.

derberg:
**Description**

- add support for tags  externalDocs url and tags in general in messages
- added one additional if in tag macro to not get empty div in case of no tags
- added some development help

I choose to not use externalDocs.description as it would complicate first iteration a lot. There are too many combinations available in case if we want to support `externalDocs.description` because we would have to take into account that user would like to use `description` like `Click here for more info` to the tag could not be clickable link but somehow the text inside tooltip.

Now, in case `url` is provided, the `name` of the tag becomes a hyperlink as discussed in the issue.

To test it with your spec just run `ag <your-asyncapi-file> https://github.com/derberg/html-template#support-tags`

Screenshot 2020-04-15 at 15 06 16

**Related issue(s)**
Fixes #6 and #5

### DIFF
--- a/partials/tags.html
+++ b/partials/tags.html
@@ -3,7 +3,7 @@
 <div class="mt-4">
   {% for tag in tagList %}
       {% if tag.externalDocs() %}
-        <span class="bg-grey-dark font-normal text-sm no-underline text-white rounded lowercase mr-2 px-2 py-1" title="{{tag.description()}}"><a href="{{tag.externalDocs().url()}}" target="_blank">{{tag.name()}}</a></span>
+        <a href="{{tag.externalDocs().url()}}" target="_blank" class="bg-grey-dark font-normal text-sm text-white rounded lowercase mr-2 px-2 py-1" title="{{tag.description()}}">{{tag.name()}}</a>
       {% else %}
         <span class="bg-grey-dark font-normal text-sm no-underline text-white rounded lowercase mr-2 px-2 py-1" title="{{tag.description()}}">{{tag.name()}}</span>
       {% endif %}
--- a/partials/tags.html
+++ b/partials/tags.html
@@ -1,8 +1,14 @@
 {% macro tags(tagList) %}
+{% if tagList %}
 <div class="mt-4">
   {% for tag in tagList %}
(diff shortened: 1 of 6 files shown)


naderafshari:
It would be great to get this in master after @fmvilas reviews. Thank you all

partials/tags.html
```diff
@@ -1,8 +1,14 @@
 {% macro tags(tagList) %}
+{% if tagList %}
 <div class="mt-4">
   {% for tag in tagList %}
-    <span class="bg-grey-dark font-normal text-sm no-underline text-white rounded lowercase mr-2 px-2 py-1" title="{{tag.description()}}">{{tag.name()}}</span>
+      {% if tag.externalDocs() %}
+        <span class="bg-grey-dark font-normal text-sm no-underline text-white rounded lowercase mr-2 px-2 py-1" title="{{tag.description()}}"><a href="{{tag.externalDocs().url()}}" target="_blank">{{tag.name()}}</a></span>
```
fmvilas, line 6: Just a quick thing: this results in black over dark-grey background. The contrast is really low. Can we please choose another color for the text? Maybe just white? Also, the span tag here is unnecessary. Replace it with the `a` tag straight ahead.

```suggestion
        <a href="{{tag.externalDocs().url()}}" target="_blank" class="bg-grey-dark font-normal text-sm no-underline text-white rounded lowercase mr-2 px-2 py-1" title="{{tag.description()}}">{{tag.name()}}</a>
```

fmvilas: I just realized that in this case you may want to leave the underline to indicate it's a link:

```suggestion
        <a href="{{tag.externalDocs().url()}}" target="_blank" class="bg-grey-dark font-normal text-sm text-white rounded lowercase mr-2 px-2 py-1" title="{{tag.description()}}">{{tag.name()}}</a>
```

derberg:
Oh my, me picking colors 😨 



How about what I just pushed:

Screenshot 2020-04-16 at 11 29 46

PR #10, opened 2020-04-17
feat: use new generic filters

derberg:
**Description**

- use generic filters library
- modify and cleanup of local filters for new testable style
- adjust template to use new filters

**Related issue(s)**
See also https://github.com/asyncapi/generator/issues/212

### DIFF
--- a/filters/all.js
+++ b/filters/all.js
@@ -1,55 +1,84 @@
-const _ = require('lodash');
-const Markdown = require('markdown-it');
-const OpenAPISampler = require('openapi-sampler');
+const filter = module.exports;
 
-module.exports = ({ Nunjucks }) => {
-  Nunjucks.addFilter('split', (string, separator) => {
-    if (typeof string !== 'string') return string;
-    const regex = new RegExp(separator, 'g');
-    return string.split(regex);
-  });
+function isExpandable(obj) {
+  if (
+    obj.type() === 'object' ||
+    obj.type() === 'array' ||
+    (obj.oneOf() && obj.oneOf().length) ||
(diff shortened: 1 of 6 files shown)


filters/all.js
```diff
@@ -1,55 +1,84 @@
...
+    obj.additionalItems() ||
+    (obj.properties() && Object.keys(obj.properties()).length) ||
+    obj.additionalProperties() ||
+    (obj.extensions() && Object.keys(obj.extensions()).filter(e => !e.startsWith('x-parser-')).length) ||
+    obj.patternProperties()
+  ) return true;
+
+  return false;
+}
+filter.isExpandable = isExpandable;
```
fmvilas, line 20: Same as in the other PR. We can probably rewrite this as follows:

```suggestion
filter.isExpandable = function(obj) {
  if (
    obj.type() === 'object' ||
    obj.type() === 'array' ||
    (obj.oneOf() && obj.oneOf().length) ||
    (obj.anyOf() && obj.anyOf().length) ||
    (obj.allOf() && obj.allOf().length) ||
    obj.items() ||
    obj.additionalItems() ||
    (obj.properties() && Object.keys(obj.properties()).length) ||
    obj.additionalProperties() ||
    (obj.extensions() && Object.keys(obj.extensions()).filter(e => !e.startsWith('x-parser-')).length) ||
    obj.patternProperties()
  ) return true;

  return false;
};
```

derberg: here I can do it this way, unless we would like to publish API.md for template filters too?

fmvilas: Yeah, API.md for filters would be great too. Leave it as is so we can add docs later 👍

PR #16, opened 2020-04-26
feat: flatten allOf cases in message payloads 

### LINKED ISSUE #11: allOf rendered in an unexpected way
#### Describe the bug

I'm not sure if this is a bug or not, but for me, allOf is rendered in an unexpected way when used in the message payload. Can you confirm if this is deliberate or not? If not, I'll look to fix.

#### How to Reproduce

```
asyncapi: 2.0.0
info:
  title: MQTT API
  version: '2.0.0'

defaultContentType: application/json

channels:
  /v2/abc/data/pack_data:
    description: Publish logging data from devices
    publish:
      operationId: publishPackData
      message:
        name: pack_data
        payload:
          $ref: "#/components/schemas/pack_data"

components:
  schemas:
    pack_data:
      allOf:
      - $ref: "#/components/schemas/common_payload"
      - type: object
        properties:
          cycle_id:
            type: string
    common_payload:
      type: object
      properties:
        device_id:
            type: string
        site_id:
            type: string
```

Screenshot 2020-04-23 at 13 47 24
#### Expected behavior

I'd expect the use of 'allOf' to be transparent in the documentation (ie as if everything had just been defined explicitly in-line).

jamescrowley:
**Description**

This PR flattens usage of 'allOf' in message payloads. Usage of 'additionalProperties' can be problematic when using allOf, as it requires *all* provided schemas to successfully validate a message. See https://json-schema.org/understanding-json-schema/reference/combining.html#allof. 

Therefore, most scenarios other than when 'additionalProperties' are permitted in every schema are likely to be edge cases as far as I can tell. Feedback especially welcome on this issue.

*Note*: I have based this off the changes in https://github.com/asyncapi/html-template/pull/13 for now

**Related issue(s)**

(Attempts) to Fix #11

### DIFF
--- a/filters/all.js
+++ b/filters/all.js
@@ -107,7 +107,10 @@ module.exports = ({ Nunjucks }) => {
   Nunjucks.addFilter('isArray', (arr) => {
     return Array.isArray(arr);
   });
-
+  
+  Nunjucks.addFilter('additionalPropertiesAllowedForAll', (items) => items.every(p => p.additionalProperties() === true || p.additionalProperties() === undefined || p.additionalProperties() === null));
+  Nunjucks.addFilter('additionalPropertiesDisabledForAll', (items) => items.every(p => p.additionalProperties() === false) );
+  
   Nunjucks.addFilter('isObject', (obj) => {
     return typeof obj === 'object' && obj !== null;
   });
--- a/partials/schema-prop.html
+++ b/partials/schema-prop.html
@@ -194,6 +194,26 @@
               {{ schemaProp(pChild, pName, odd=(not odd), required=(p.required() | contains(pName))) }}
             {% endfor %}
           {% endfor %}
+          {# allOf requires *all* provided schemas to successfully validate a message
+             and so 'additionalProperties' can be problematic.
+             See https://json-schema.org/understanding-json-schema/reference/combining.html#allof
+             Therefore, I suspect most scenarios other than the one where additionalProperties 
+             are allowed in every schema are likely to be edge cases
+          #}
+          {% if prop.allOf() | additionalPropertiesAllowedForAll %}
+            <p class="pl-6 mb-2 mt-4 text-xs text-grey-darker">Additional properties are allowed.</p>
+          {% elif prop.allOf() | additionalPropertiesDisabledForAll %}
+            <p class="pl-6 mb-2 mt-4 text-xs text-grey-darker">Additional properties are <strong>NOT</strong> allowed.</p>
+          {% else %}
+            <p class="pl-6 mb-2 mt-4 text-xs font-bold text-grey-darker">Additional properties must adhere to the following schema:</p>
+            {% for p in prop.allOf() %}
+              {% if p.additionalProperties() | isObject %}
+                <div class="{% if odd %}bg-grey-lightest{% else %}bg-grey-lighter{% endif %} pl-6">
+                  {{ schemaProp(p.additionalProperties(), '(property name)', odd=(not odd), open=false, specialName=true) }}
+                </div>
+              {% endif %}
+            {% endfor %}
+          {% endif %}
         {% endif %}
       </div>
     {% endif %}
--- a/partials/schema-prop.html
+++ b/partials/schema-prop.html
@@ -5,6 +5,7 @@
 <div class="{% if odd %}bg-grey-lighter{% else %}bg-grey-lightest{% endif %} {% if not root %}pl-8 pr-8{% endif %} rounded">
   <div class="{% if open %}is-open{% endif %}">
     <div class="{% if prop | isExpandable %}js-prop cursor-pointer py-2{% endif %} flex property">
+      {% if (not prop.allOf()) %}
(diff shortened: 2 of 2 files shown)


filters/all.js
```diff
@@ -107,7 +107,10 @@ module.exports = ({ Nunjucks }) => {
   Nunjucks.addFilter('isArray', (arr) => {
     return Array.isArray(arr);
   });
-
+  
+  Nunjucks.addFilter('additionalPropertiesAllowedForAll', (items) => items.every(p => p.additionalProperties() === true || p.additionalProperties() === undefined || p.additionalProperties() === null));
+  Nunjucks.addFilter('additionalPropertiesDisabledForAll', (items) => items.every(p => p.additionalProperties() === false) );
```
jamescrowley, line 112: I couldn't find a nice way to in-line this within the template. Suggestions welcome!

test/spec/asyncapi.yml
```diff
@@ -145,6 +145,7 @@ components:
           x-pi: false
         sentAt:
           $ref: "#/components/schemas/sentAt"
+      additionalProperties: false
```
jamescrowley, line 148: There are currently no allOf usages in the template. I wasn't sure on the best example to add here.

derberg:
@jamescrowley How should I treat https://github.com/asyncapi/html-template/pull/13/files? as it is correct and I can merge, but you also did the same changes in this PR

jamescrowley:
@derberg I based this PR off that one, so if you're happy with https://github.com/asyncapi/html-template/pull/13/ merge that first and I'll rebase - and if you're not, then I'll remove the changes :)

derberg:
@jamescrowley lgtm, I just adjusted the PR summary as we use conventional commits. You can rebase this PR

jamescrowley:
@derberg I've re-based. I'm leaning towards just removing the 'additionalProperties' display entirely, given it's not really clear what the author's intent would be? While technically they might be permitted (if additionalProperties is missing or set to true), but that could just be due to the json schema limitations. Thoughts?

derberg:
@jamescrowley I merged into `master` some refactor of filters, you will have to rebase and as a result modify the filters a bit.

Also please extend `asyncapi.yml` with `allOf` use case as we always want to have a new feature represented in the example. I wrote about it here https://github.com/asyncapi/html-template/pull/16#discussion_r416107543

derberg:
I also played with the template a bit. This is the spec I was using:
```
asyncapi: 2.0.0

info:
  title: Signup service example (internal)
  version: 0.1.0

channels:
  /user/signedup:
    subscribe:
      message:
        payload:
          $ref: "#/components/schemas/Cat"

components:
  schemas:
    Pet:
      type: object
      discriminator: petType
      properties:
        name:
          type: string
        petType:
          type: string
      required:
      - name
      - petType
    Cat:
      description: A representation of a cat
      allOf:
      - $ref: '#/components/schemas/Pet'
      - type: object
        properties:
          huntingSkill:
            type: string
            description: The measured skill for hunting
            enum:
            - clueless
            - lazy
            - adventurous
            - aggressive
        required:
        - huntingSkill
    Dog:
      description: A representation of a dog
      allOf:
      - $ref: '#/components/schemas/Pet'
      - type: object
        properties:
          packSize:
            type: integer
            format: int32
            description: the size of the pack the dog is from
            minimum: 0
        required:
        - packSize
    StickInsect:
      description: A representation of an Australian walking stick
      allOf:
      - $ref: '#/components/schemas/Pet'
      - type: object
        properties:
          petType:
            const: StickBug
          color:
            type: string
        required:
        - color
```

This is how it looks like with current template:
Screenshot 2020-04-29 at 17 57 44

This is how it looks like with your PR:
Screenshot 2020-04-29 at 17 54 52



1. You can see that below is missing in your PR, we shoudl be able to collapse/expande object payload
Screenshot 2020-04-29 at 18 05 21
2. I'm also super puzzled about one thing. It is not clear for the user that the given payload is `allOF` anymore and I'm thinking if it is not bad, because the payload part should reflect the reality and the example should be flattened as it is not. Thoughts?

fmvilas:
> I'm also super puzzled about one thing. It is not clear for the user that the given payload is `allOF` anymore and I'm thinking if it is not bad, because the payload part should reflect the reality and the example should be flattened as it is not. Thoughts? 

I think this is a valid concern. It could probably be configured using a template param? So people can decide what to do. For instance, for public APIs, you might not want to share your AsyncAPI files. However, if it's something internal and you let people inspect the AsyncAPI file, it's useful to have an indicator that you're using inheritance.

We can also have just an indicator nearby the schema using `allOf`. Just remember not only the payload may be using `allOf`.

jamescrowley:
There are other problems here unfortunately. Within allOf, you can define the same properties in each definition, with different levels of 'specificity' on the constraint (ie, maybe just type: object, and another schema one putting tighter restrictions on it). Right now, the template just ends up listing the properties twice. In theory you could try and merge this definitions into the most constrained version, but we're definitely veering into complex territory. It may be worth consulting closer with how OpenAPI approaches this.

PR #17, opened 2020-04-28
feat: Flag deprecated properties

jamescrowley:
**Description**

Now that https://github.com/asyncapi/parser-js/issues/78#issuecomment-620469576 has been resolved, and deprecated() is available, flag properties that have the deprecated attribute.

**Note** I'm not sure how to show that we now have a dependency on parser-js > 0.18.3?

```
cb_hardware_version:
    type: string
    deprecated: true
```

is displayed as:

Screenshot 2020-04-28 at 18 53 29

PR #18, opened 2020-04-28
feat: add generator config param

fmvilas:
**Description**
Adds `generator` to the config file to indicate the version-range it's compatible with. The reason I'm choosing this minimum version is this PR: https://github.com/asyncapi/html-template/pull/17.

**Related issue(s)**
Partially solves https://github.com/asyncapi/generator/issues/280

### DIFF
--- a/.tp-config.json
+++ b/.tp-config.json
@@ -8,5 +8,6 @@
       "description": "Sets the base URL for links and forms.",
       "required": false
     }
-  }
+  },
+  "generator": "<2.0.0"
 }
--- a/.tp-config.json
+++ b/.tp-config.json
@@ -9,5 +9,5 @@
       "required": false
     }
   },
-  "generator": "<2.0.0"
+  "generator": ">=0.40.1 <2.0.0"
 }


.tp-config.json
```diff
@@ -8,5 +8,6 @@
       "description": "Sets the base URL for links and forms.",
       "required": false
     }
-  }
+  },
+  "generator": ">=0.40.1 <2.0.0"
```
Tenischev, line 12: I doubt `<2.0.0`. This means that the template will work flawlessly on the next major version, despite some breaking changes that may be made to it.

fmvilas: It means that this template will work with all the generator@0.40.1 and above and all the generator@1.x. There shouldn't be breaking changes affecting the template but if there are at some point we can always change this parameter. Actually, at some point after releasing v1 of Generator, we'll change this to `^1.0.0`.

Tenischev: Yeah, i mean exactly this, that you claim support of `generator@1.x.y` which is not even released :smile: 
And by [semver ](https://semver.org/#summary) - `MAJOR version when you make incompatible API changes`. So, for me it's rather bold to speak that template will support next generator version...

fmvilas: Yeah, I should clarify that the work we're doing right now is precisely to launch v1 in the next week. That's why I'm already putting it there.

As per [semver](https://semver.org/#spec-item-4) too: `Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.` and `Version 1.0.0 defines the public API. The way in which the version number is incremented after this release is dependent on this public API and how it changes.`.

So there's no need to make a breaking change to go from major 0 to 1. You jump when you think it's stable.

fmvilas: Also, the version of this package is still on 0.x. Meaning any new version might contain breaking changes.

Tenischev: > launch v1 in the next week.

That's great, no concerns then

derberg: > launch v1 in the next week.

@Tenischev I'm pretty sure @fmvilas ment `weeks` 😅

fmvilas: Yes! 😄

PR #31, opened 2020-05-22
feat: list all messages in sidebar

### LINKED ISSUE #29: list all messages in side bar
We have a lot of messages under one channel. It's good to have list of all messages, linked to there reference in the schema. We had this available in the old HTML template, but not in the new one.

image

- What changes have to be introduced: List all messages after listing channels
- Will this be a breaking change: No
- How could it be implemented/designed: I have a PR

WaleedAshraf:
closes #29

Updated view:
image

### DIFF
--- a/partials/content.html
+++ b/partials/content.html
@@ -5,3 +5,4 @@
 {% if asyncapi.hasChannels() %}
 {% include "./operations.html" %}
 {% endif %}
+{% include "./messages.html" %}
--- a/partials/message.html
+++ b/partials/message.html
@@ -3,7 +3,6 @@
 
 {% macro message(msg, showIndex=false, index=0, open=false) %}
 
-<a name="message-{{msg.uid()}}"></a>
 <div class="bg-grey-lighter rounded p-4 mt-2">
   <div class="text-sm text-grey-darker mb-2">
     {% if showIndex %}
--- a/partials/message.html
+++ b/partials/message.html
@@ -3,6 +3,7 @@
 
 {% macro message(msg, showIndex=false, index=0, open=false) %}
 
+<a name="message-{{msg.uid()}}"></a>
 <div class="bg-grey-lighter rounded p-4 mt-2">
   <div class="text-sm text-grey-darker mb-2">
     {% if showIndex %}
--- a/partials/messages.html
+++ b/partials/messages.html
@@ -1,0 +1,15 @@
+{% from "./message.html" import message %}
+
+<a name="messages"></a>
+<h2 class="ml-8 mt-4">Messages</h2>
+
+<div class="responsive-container">
+  <div class="center-block p-8">
+    <div class="pb-8">
+      {% for key, value in asyncapi.allMessages() %}
+      <a name="message-{{key}}"></a>
+      {{ message(value, showIndex=true, index=loop.index, open=false) }}
+      {% endfor %}
+    </div>
+  </div>
+</div>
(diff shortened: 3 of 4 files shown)


derberg:
@WaleedAshraf and we also use conventional commits to drive releases automation https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md#conventional-commits
I already changed the PR title for you

derberg:
I'm wondering if this is not something that should be behind a parameter so users can decide if they want it or not, but yeah, let us add it by default and see what community has to say on it

derberg:
@WaleedAshraf one more question, you are aware that one message can be used in multiple channels and as a result, like in the example, link like `<a name="message-turnOnOff"></a>` gets duplicated, and message link will work only with a first occurrence in the list. Shouldn't there messages be rendered separately like here https://asyncapi.github.io/asyncapi-react/ ? and links from navigation point to this separate section with messages only?

WaleedAshraf:
@bpedro you are right. There can be multiple links for one message, and this fix will only take you to the first link.
My understanding is that in the future we can have filters. And when someone clicks a message, we apply a filter and only show messages which have that schema.

Using filters can also be useful for tags. i.e, we list all the tags, and clicking on one tag will filter all messages which have that tag. etc.

The way you showed above will also work. For that first, we need to 
- make schema collapses able
- render messages, servers, schema in separate sections

derberg:
@WaleedAshraf this filter functionality if pretty sophisticated, do you plan to work on it?

with this PR schema doesn't have to be collapsable, and you would only have to render messages (servers or schemas are not part of this PR) below operations, and link to messages there only where you do not have duplicates, instead of a message in a given operation.

WaleedAshraf:
Ok, I'll work on rending messages in below Operation and linking it there. 👍

WaleedAshraf:
@derberg 

Kapture 2020-06-01 at 19 31 27

partials/content.html
```diff
@@ -5,3 +5,4 @@
 {% if asyncapi.hasChannels() %}
 {% include "./operations.html" %}
 {% endif %}
+{% include "./messages.html" %}
```
WaleedAshraf, line 8: It would be good to have `.hasMessages()` check here. But it doesn't exist now.
We can add that in the future.

derberg: Please create an issue in parser to add it so we don't forget

derberg:
@WaleedAshraf lgtm, just create an issue for `.hasMessage()`
I guess people might want to have examples next to those messages but yeah, let community request to make sure they need it

rv0:
> I'm wondering if this is not something that should be behind a parameter so users can decide if they want it or not, but yeah, let us add it by default and see what community has to say on it

oh boy :) 
our api is quite extensive and most payloads aren't named (sometimes we should and eventually will, but usually it does not make sense).. so currently we have like 250+ "<anonymous-message-xyz>" entries shown in the sidebar

In our case, having these entries in the sidebar offers no added value, even if they were all named, there's simply way too much of them to have any use.

derberg:
@rv0 thanks for sharing, we have use case, makes sense to do it. Please create a separate issue. Adding param is easy, we just need to agree on how to name it and what is the default.

btw do you know that you can lock for a specific `html-template` version so you are not surprised with new features you might not want to have? :)

rv0:
@derberg will create issue eventually, I'm aware of the version targeting, it's great how fast this project and it's subprojects are advancing
it's just a bit weird that features are released without a way to configure them (as you say yourself, adding param is easy), hence my reply here

derberg:
@rv0 please keep in mind that this template is not yet on 1.0 release hence the approach is more relaxed here. It would be great though to work together on some design document that outlines what features should be parametrized and what not, or maybe really all.

Issue #23, opened 2020-05-13
Fold long examples


derberg:
@Tamriel commented on [Wed Jul 17 2019](https://github.com/asyncapi/generator/issues/72)

When you have long examples, they take up so much space in the right column, that in the middle column there will be very much white space between one operation to the next.

The OpenApi html generator [ReDoc](https://redocly.github.io/redoc/#operation/findPetsByStatus) solves this by folding the examples. The user is able to expand and collapse them.

---

@fmvilas commented on [Fri Jul 19 2019](https://github.com/asyncapi/generator/issues/72#issuecomment-513027846)

Absolutely agree.

---

@jonaslagoni commented on [Fri Nov 08 2019](https://github.com/asyncapi/generator/issues/72#issuecomment-552005320)

I think this was the same as #126 and closed by #153

---

@fmvilas commented on [Sat Nov 09 2019](https://github.com/asyncapi/generator/issues/72#issuecomment-552084458)

🤔 I don't think it's the same. The problem here is that, when the example is so long, it takes a lot of space on the screen. We're talking about limiting the visible height of the example and having an Expand/Collapse link/button below the example.

---

@derberg commented on [Sat Feb 15 2020](https://github.com/asyncapi/generator/issues/72#issuecomment-586627367)

It would be great to get some sample asyncapi spec that is actually generating such a long content, or do you know already some good public example?

Regarding ReDoc. Can you actually point me to a section where this functionality is visible? Is it expand/collapse by default to all, or only for the ones that are longer than the middle section?

In asyncapi-react component, it is solved in the way that all is collapsed by default, and you have one button to expand all, or you can expand one-by-one. So there are different ways to do it, this is why I ask :)

---

@github-actions[bot] commented on [Wed May 13 2020](https://github.com/asyncapi/generator/issues/72#issuecomment-627664607)

This issue has been automatically marked as stale because it has not had recent activity :sleeping:
It will be closed in 30 days if no further activity occurs. To unstale this issue, add a comment with detailed explanation. 
Thank you for your contributions :heart:

WaleedAshraf:
@derberg check here for an example: https://redocly.github.io/redoc/#tag/pet
I'll try a take on this.

Issue #19, opened 2020-04-28
[HTML template] Ability to auto-expand message schema


derberg:
@davydog187 commented on [Wed Jan 22 2020](https://github.com/asyncapi/generator/issues/202)

Hello and thank you for this great project!

# The Problem
I have nested payload structures that are 3-4 levels deep, and it is cumbersome to keep clicking to view the schema. 

# Potential Solution
It would be nice if there was an option in the generator that has nested structures auto-expanded, maybe with a parameter to specify the level of nesting that you want to stop auto-expanding. 

I could be interested in helping contribute, but would need some pointing in the right direction :) Thanks!

---

@fmvilas commented on [Wed Feb 05 2020](https://github.com/asyncapi/generator/issues/202#issuecomment-582338627)

Hey @davydog187. Thanks for the suggestion. It would indeed be great. Any help would be much appreciated 🙏

naderafshari:
Hi All @here 
This issue is very important to us as we have a number of messages with a large set of fields and nested and complex schema. Browsing through the html page going from one message to the next, not so practical!

I believe the react component (https://github.com/asyncapi/asyncapi-react) has this feature already. It would be great that html-template follows suit.

Also in each message there is an example for the message. It would be great for that section have the ability to expand and collapse. So maybe the example is collapsed by default and can be expanded when user is interested in seeing the example.

Just to give an idea, here is the SMALLEST message we have...
`ConfigChngEvnt:
  name: ConfigChngEvntMessage
  description: This message is used for notifying configuration changes on the controlbus.
  bindings:
    amqp:
      contentEncoding: gzip
      messageType: 'user.signup'
      bindingVersion: 0.1.0
  tags:  
    - name: correlationId
      description: click for more information on correlationId 
    - name: subject
      description: click for more information on correlationId 
  payload:
    type: object
    properties:
      # Event Header
      specverson:
        $ref: ../../common/yml/Specversion.yml
      type:
        $ref: ../../common/yml/Type.yml
      source:
        $ref: ../../common/yml/Source.yml
      id:
        $ref: ../../common/yml/Id.yml
      correlationId:
        $ref: ../../common/yml/CorrelationId.yml
      subject: 
        $ref: ../../common/yml/Subject.yml
      eventTime:
        $ref: ../../common/yml/TimestampHub.yml
      datacontenttype:
        $ref: ../../common/yml/Datacontenttype.yml
      dataschema:
        $ref: ../../common/yml/Dataschema.yml

      data:
[...]

PR #34, opened 2020-05-22
feat: make payload example collapsible

WaleedAshraf:
Possible fix for #23 

I need help from someone.

Kapture 2020-05-22 at 14 22 20

### DIFF
--- a/partials/example.html
+++ b/partials/example.html
@@ -1,8 +1,8 @@
 {% macro example(msg, channelName, open=false) %}
 
 <form>
-  <h5 class="examples-uid">{{msg.uid()}}</h5>
-  <div class="payload-examples mt-4 {% if open %}is-open{% endif %}">
+  <h5 class="examples-uid text-orange mt-4">{{msg.uid()}}</h5>
+  <div class="payload-examples mb-4 {% if open %}is-open{% endif %}">
     <div class="{% if  msg.payload() | isExpandable %}js-prop cursor-pointer {% endif %} flex property">
       <span class="px-2 mr-2 text-grey-lightest text-sm border rounded focus:outline-none cursor-pointer {% if not msg.payload() %}text-grey-dark border-grey-dark opacity-25{% else %}text-grey-lightest{% endif %}" {% if not msg.payload() %}title="This message does not have a payload"{% endif %}>Payload</span>
       {% if msg.payload() | isExpandable %}
@@ -25,7 +25,7 @@
     </div>
   </div>
 
-  <div class="headers-examples mt-4 {% if open %}is-open{% endif %}">
+  <div class="headers-examples mb-4 {% if open %}is-open{% endif %}">
     <div class="{% if  msg.payload() | isExpandable %}js-prop cursor-pointer {% endif %} flex property">
       <span class="px-2 mr-2 text-sm border rounded focus:outline-none cursor-pointer {% if not msg.headers() %}text-grey-dark border-grey-dark opacity-25{% else %}text-grey-lightest{% endif %}" {% if not msg.headers() %}title="This message does not have headers"{% endif %}>Headers</span>
       {% if msg.payload() | isExpandable %}
--- a/partials/example.html
+++ b/partials/example.html
@@ -2,51 +2,55 @@
 
 <form>
   <h5 class="examples-uid text-orange mt-4">{{msg.uid()}}</h5>
-  <div class="payload-examples mb-4 {% if open %}is-open{% endif %}">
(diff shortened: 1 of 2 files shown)


derberg:
@WaleedAshraf what help you need? I like how it is done in ReDoc

derberg:
@WaleedAshraf there's also a duplicate issue here https://github.com/asyncapi/html-template/issues/19#issuecomment-620811026 with a good example of a big payload

WaleedAshraf:
The way it is done in reDoc is too much work for me to do. I couldn't understand how they do it in the back.

This fix only solves the problem by making an example collapsable. We can later improve it to make it like how redoc has done it. 

Help needed: right now, it's showing both examples (headers + payload) in one view, it should be separate in two views and both should be collapsable. This is the point where my CSS skills ends.

derberg:
@WaleedAshraf ok, I get ya, current `>` looks strange and is not that visible, reDoc clearly says Expand/Collaps so it would make more sense to have it clear on our side to, so instead the chevron icon I would suggest a clickable label Expand Example that once clicked, changes to Collapse Example. Or have a text/link and a chevron :) 

Regarding problem that header is visible in payload, and the other way around as well.
It is because you no longer use classes like `payload-examples mt-4` and `headers-examples mt-4`
 
Screenshot 2020-05-26 at 12 09 13

WaleedAshraf:
redDoc have ExpandAll/CollapseAll because you can expand individual object.

In our case, we don't have that option.
Also, we use the only chevron all over the place. In schema, payload etc.

Shouldn't we keep it consistent?

derberg:
good point about consistency, but then should instead of
Screenshot 2020-05-26 at 16 13 41

be something we already have, but with examples
Screenshot 2020-05-26 at 16 13 50

chevron is not that bad, but for the given background the color should be different, white?

fmvilas:
IMO chevron is good (only needs to be white for contrast reasons, as Lukasz pointed out) but it's really weird there's no text along with it so I'd add something like `Example #0 >`.

WaleedAshraf:
@fmvilas Yes, I'll the message name also with example so that examples don't mix up.

WaleedAshraf:
@fmvilas @derberg how is this.

Kapture 2020-05-27 at 12 26 02

derberg:
@WaleedAshraf message name is useful in case there are multiple messages at a channel, makes sense. What @fmvilas meant is that it is weird that the chevron is there alone, with no label. As I wrote here https://github.com/asyncapi/html-template/pull/34#issuecomment-634053761, maybe instead of those Payload and Header buttons, there should be just:
```
Examples

lightMeasured
Payload >
Headers >

turnOnOff
Payload >
Headers >
```
buttons look nice but using them looks cumbersome, advantage of my proposal is that you can expand both, payload and headers and see both at the same time

WaleedAshraf:
@derberg 
Kapture 2020-05-27 at 13 52 31

derberg:
@WaleedAshraf I'm in with my whole ❤️ 

would be awesome to see how it works with very large example.
@naderafshari could you check it with your spec documents? you can quickly test it with latest generator `0.50.1` and instead of using `@asyncapi/html-template` just put `https://github.com/asyncapi/html-template#issue-23-fix`

WaleedAshraf:
need to apply few more fixes... in progress...

WaleedAshraf:
@derberg ready for testing

Kapture 2020-05-27 at 16 04 04

derberg:
@WaleedAshraf I checked against this https://github.com/metricq/metricq-rpc-docs/blob/master/manager.asyncapi.yaml that uses oneOf and there is some weird behavior

Screenshot 2020-05-28 at 10 19 19


Also, I believe that the Headers button should not be visible if there are no headers. I know that in current version it is grayed out and you left it like this, but in your new version there is also chevron, that you can still click to expand and collapse, even if headers are not there.  I hope you know what I mean 😄 so in the version you work on I vote for not having headers at all if they are not present

WaleedAshraf:
https://github.com/metricq/metricq-rpc-docs/blob/master/manager.asyncapi.yaml 
This is working fine for me.
Check now, payload and headers button is hidden when not available.

Kapture 2020-05-28 at 10 34 36

derberg:
@WaleedAshraf lgtm, I bet people will request a button to expand/collapse all examples with one click.

WaleedAshraf:
they are welcome to open a pr. ;)

derberg:
@all-contributors please add @WaleedAshraf for code, bug

PR #40, opened 2020-06-02
refactor: replace msg.name() with msg.uid()

### LINKED ISSUE #39: Use message.uid() rather than optional message.name()
#### Reason/Context
More a cleanup proposal/discussion towards how message names are rendered in the resulting HTML according to [this PR](https://github.com/asyncapi/markdown-template/pull/28) in the `markdown-template`.

#### Description

https://github.com/asyncapi/html-template/blob/03eb7dd360d19f85bd143bd7ef643c8d11e04496/partials/message.html#L11-L20

This block would need to be changed to use `msg.uid()`. When to actually include it should be discussed. With the mentioned PR in `markdown-template`, using the `name` property in messages is optional since the template falls back to the property-name of the message-object. Would be nice to see a similar behavior in the HTML template as well.

schw4rzlicht:
**Description**
Replaced `msg.name()` with `msg.uid()`.

**Related issue(s)**
Closes #39

### DIFF
--- a/partials/message.html
+++ b/partials/message.html
@@ -10,13 +10,9 @@
     {% endif %}
     {% if msg.title() %}
       {{msg.title()}}
-      {% if msg.name() %}
-        <span class="border text-orange rounded text-xs ml-3 py-0 px-2">{{msg.name()}}</span>
-      {% endif %}
+      <span class="border text-orange rounded text-xs ml-3 py-0 px-2">{{msg.uid()}}</span>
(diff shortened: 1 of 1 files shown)


derberg:
@schw4rzlicht I only changed the commit message to indicate it is not a feature or fix to not trigger release, as this is rather refactor that improves code

derberg:
@all-contributors please add @schw4rzlicht for code

Issue #42, opened 2020-06-04
Sidebar partial use document root tags, not operation tags


sebastian-palma:
Currently the sidebar is iterating over each element in `asyncapi.tags()` and checking if that tag name is in the channel tags (publish/subscribe) when they have tags:

```html
{% if channel.publish() and channel.publish().hasTags() and channel.publish() | containTags(categorytags) %}
...
{% endif %}

{% if channel.subscribe() and channel.subscribe().hasTags() and channel.subscribe() | containTags(categorytags) %}
...
{% endif %}
```

But that means you need strictly to have the same tags you have in every channel, at the root of the document. Otherwise, they're all rendered as Untagged.

In an example code like [this](https://gist.github.com/sebastian-palma/eeb128d38b2769d7ce12f8f033b0da15) (see both `foo` and `bar` tags in `smartylighting/streetlights/1/0/event/{streetlightId}/lighting/measured` and `smartylighting/streetlights/1/0/action/{streetlightId}/turn/on` respectively), running the generator with the `sidebarOrganization` equals to `byTags` would generate the sidebar like this:

![Screen Shot 2020-06-04 at 12 55 23](https://user-images.githubusercontent.com/11888191/83743299-35095a80-a663-11ea-950c-f773188c2da3.png)

And if you add the tags to the root of the document as:

```yaml
...
tags:
  - name: foo
  - name: bar
```

The sidebar is generated like this:

![Screen Shot 2020-06-04 at 13 01 30](https://user-images.githubusercontent.com/11888191/83743610-a0532c80-a663-11ea-94c5-b0db2bfae244.png)

Foo and Bar have no channels, they're placed under the Untagged _label_.

It seems the logic of the sidebar regarding tags was written taking in mind that every tag under every channel must be in the root tags, so they can be grouped by iterating the root tags and then associating them as long as the channel tags contain the root tag from the iteration, which is okay because it works, but it'd force us to duplicate the place where we add the tags; once in the root and once for each channel having tags.

derberg:
@fmvilas do you recall why `asyncapi.tags()` was added to the logic that builds tag left nav? when I look at it, it seems to me it could never work

fmvilas:
This was the initial issue: https://github.com/asyncapi/generator/issues/69. It's another way to structure the sidebar.

sebastian-palma:
Do we consider adding a way to render tags without having them in the root of the document? Or do we close the issue in favor of keeping the current functioning @fmvilas, @derberg?

[...]

PR #44, opened 2020-06-14
feat: Add flag to use operations tags instead of root tags for the sidebar

sebastian-palma:
Implements https://github.com/asyncapi/html-template/issues/42.

`all.js` now defines a function `operationsTags` to get subscription/publish operations tags and use them to build the sidebar.

New functionality available under `sidebarOrganization` param with `byTagsNoRoot` value

### DIFF
--- a/filters/all.js
+++ b/filters/all.js
@@ -1,19 +1,20 @@
 const filter = module.exports;
 
 function isExpandable(obj) {
+  const fun = (obj) => typeof obj === "function";
   if (
-    (typeof obj.type === "function" && obj.type() === "object") ||
-    (typeof obj.type === "function" && obj.type() === "array") ||
-    (typeof obj.oneOf === "function" && obj.oneOf() && obj.oneOf().length) ||
-    (typeof obj.anyOf === "function" && obj.anyOf() && obj.anyOf().length) ||
-    (typeof obj.allOf === "function" && obj.allOf() && obj.allOf().length) ||
-    (typeof obj.items === "function" && obj.items()) ||
-    (typeof obj.additionalItems === "function" && obj.additionalItems()) ||
-    (typeof obj.properties === "function" && obj.properties() && Object.keys(obj.properties()).length) ||
-    (typeof obj.additionalProperties === "function" && obj.additionalProperties()) ||
-    (typeof obj.extensions === "function" && obj.extensions() &&
+    (fun(obj.type) && obj.type() === "object") ||
+    (fun(obj.type) && obj.type() === "array") ||
+    (fun(obj.oneOf) && obj.oneOf() && obj.oneOf().length) ||
+    (fun(obj.anyOf) && obj.anyOf() && obj.anyOf().length) ||
+    (fun(obj.allOf) && obj.allOf() && obj.allOf().length) ||
+    (fun(obj.items) && obj.items()) ||
+    (fun(obj.additionalItems) && obj.additionalItems()) ||
+    (fun(obj.properties) && obj.properties() && Object.keys(obj.properties()).length) ||
+    (fun(obj.additionalProperties) && obj.additionalProperties()) ||
+    (fun(obj.extensions) && obj.extensions() &&
       Object.keys(obj.extensions()).filter(e => !e.startsWith("x-parser-")).length) ||
-    (typeof obj.patternProperties === "function" && obj.patternProperties())
+    (fun(obj.patternProperties) && obj.patternProperties())
   ) return true;
 
   return false;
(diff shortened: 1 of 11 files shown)


derberg:
In the issue we discussed that best would be to add new parameter or I would add not that best would be to add new value for existing `sidebarOrganization`, so not only have `byTags` but also `byTagsNoRoot`?

when I check your PR with our dummy spec https://github.com/asyncapi/generator/blob/master/test/docs/dummy.yml and I remove root tags from it, it works as expected. We just loose the functionality we had before, the one based on the root tags

Screenshot 2020-06-17 at 14 15 27

sebastian-palma:
Sure @derberg , I forgot that part.

I've added a new function to `all.js` and it made me wonder; is there some way we can test the things we modify in that file? And, do you think we could test the template generation? I mean, adding a new option should change the template, how we could make sure nothing is broken?

I ask from the complete ignorance since I've never worked with Nunjucks. I don't know how you approach these things.

derberg:
@sebastian-palma filters you can easily test as they are normal js functions. Of course, it sucks that we do not have tests yet in this template but we will need them before 1.0 release. When I was working on a library with generic filters I tested most of them, but not all, so you can have a look https://github.com/asyncapi/generator-filters/tree/master/src.

For Nunjucks, yeah, this is beast 😄 we discussed the topic of testing a while ago, for html and markdown template I do not see any other option than snapshot testing. For templates that generate code, I think best is to always have a proper setup with docker compone and integration tests, I do not see any other option.

I'm not focused on Parser, but it will not take more than 2 weeks, thank my next goal is to create a github repository template that will be a template for all the templates (inception 😄 ) so it is easier to start with generator and yeah, have an answer there in the template on how to do testing 😄 it will be a lot of fun

sebastian-palma:
Hey @derberg, it took more time than I thought 😅. I added the new flag for the sidebar organization "byTagsNoRoot" and conditional include the corresponding partials.

I tested with the dummy yaml, and the byTags flag and another yaml file with this new flag and seems to work fine, anyway, I'd appreciate you could try it to see if nothing got broken.

derberg:
@sebastian-palma sorry for the delay man, had some days off, will have a look this week

test/spec/asyncapi.yml
```diff
@@ -39,16 +39,16 @@ channels:
...
+        - name: operation-tag2
           description: Description 2
           externalDocs:
             url: "https://www.asyncapi.com/"
-        - name: oparation-tag3
-        - name: oparation-tag4
+        - name: operation-tag3
+        - name: operation-tag4
           description: Description 4
         - name: message-tag5
```
derberg, line 53: can you also remove this, it also should not be here, an old mistake

filters/all.js
```diff
@@ -99,3 +104,15 @@ function containNoTag(channels, tagsToCheck) {
...
+function operationsTags(object) {
+  let tags = new Set();
+  const extractName = (tags, acc) => tags.forEach(({ name }) => acc.add(name));
+  object.channelNames().forEach(channelName => {
+    let channel = object.channel(channelName);
+    if (channel.hasPublish()) extractName(channel._json.publish.tags || [], tags);
+    if (channel.hasSubscribe()) extractName(channel._json.subscribe.tags || [], tags);
+  });
+  return Array.from(tags);
+}
```
derberg, line 117:
```suggestion
function operationsTags(object) {
  let tags = new Set();
  const extractName = (tags, acc) => tags.forEach((tag) => acc.add(tag.name()));
  object.channelNames().forEach(channelName => {
    let channel = object.channel(channelName);
    if (channel.hasPublish() && channel.publish().hasTags()) extractName(channel.publish().tags(), tags);
    if (channel.hasSubscribe() && channel.subscribe().hasTags()) extractName(channel.subscribe().tags(), tags);
  });
  return Array.from(tags);
}
```

filters/all.js
```diff
@@ -48,7 +49,9 @@ function containTags(object, tagsToCheck) {
     let found = false;
     for (let tagToCheckIndex in tagsToCheck) {
       let tagToCheck = tagsToCheck[tagToCheckIndex]._json;
-      if (tagToCheck.name === tag.name) {
+      let tagName = tag.name;
+      if ((typeof tagToCheck !== 'undefined' && tagToCheck.name === tagName) ||
```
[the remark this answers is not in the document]
sebastian-palma: Yep, shorter and better ;)

test/spec/asyncapi.yml
sebastian-palma: The `name`s you mean? 👀

derberg: entire tag

sebastian-palma: There it is. Sorry for the delay.

derberg: no worries mate, thanks for this additional change but I think I was misunderstood, sorry, by `entire tag` I meant the tag that I commented, so just
```
        - name: message-tag5
          externalDocs:  
            url: "https://www.asyncapi.com/"
```

the rest of tags should stay, sorry again

sebastian-palma: Cool, no problemo. Just fixed and squashed the commits, thanks @derberg.

derberg:
@all-contributors please add @sebastian-palma for code

PR #45, opened 2020-07-05
chore: add all filters tests

sebastian-palma:
After adding some new code to `all.js` and as per @derberg proposed I thought it'd be better to add some tests in order to not break the existing features.

I just incorporated "Ava" as it's the one used in the generator-filters repository, which seems to work well.

The changes made to `isExpandable` (`(typeof obj.type === "function" && obj.type() === "object")`) are to first check if the object has defined a `type` function, otherwise, it throws an error. But it seems that there are some other parts of an AsyncApi object that might not have `type` defined but some of the other functions that are there checked.

### DIFF
--- a/.github/workflows/pull-request-testing.yml
+++ b/.github/workflows/pull-request-testing.yml
@@ -1,0 +1,22 @@
+name: Pull request testing
+
+on:
+  pull_request
+
+jobs:
+  release:
+    name: 'Run linter and tests'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v2
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 13
+      - name: Install dependencies
+        run: npm ci
+      - name: Run linter
+        run: npm run lint
+      - name: Run tests
+        run: npm test
--- a/filters/all.js
+++ b/filters/all.js
@@ -2,17 +2,18 @@ const filter = module.exports;
 
 function isExpandable(obj) {
   if (
-    obj.type() === 'object' ||
-    obj.type() === 'array' ||
-    (obj.oneOf() && obj.oneOf().length) ||
-    (obj.anyOf() && obj.anyOf().length) ||
-    (obj.allOf() && obj.allOf().length) ||
-    obj.items() ||
-    obj.additionalItems() ||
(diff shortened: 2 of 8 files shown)


filters/all.contain_no_tag.test.js
```diff
@@ -0,0 +1,60 @@
+const test = require("ava");
+const {containNoTag} = require("./all");
+
+let undef;
```
derberg, line 4: this should be `const` as you never modify it. It is also in other test files

filters/all.non_parser_extensions.test.js
```diff
@@ -0,0 +1,39 @@
...
+
+test("returns an empty Map when schema extensions is not a function", t => {
+  t.context.doc.extensions = [];
+  const result = nonParserExtensions(t.context.doc);
+  t.is(result.constructor.name, "Map");
+  t.is(result.size, 0);
+});
+
+test("returns a Map with the truthy schema extensions that do not start with x-parser-", t => {
+  t.context.doc.extensions = function() {
```
derberg, line 28: not super important, but I just noticed it so writing about it, could be arrow function here

sebastian-palma:
Hi @derberg thanks for checking that. I just opted for moving the tests to a separate folder, out of imagination I just named it "tests".

I placed the tests in the same filters folder before because we already have a test folder and to avoid confusion.

Now it seems to be working fine, would you mind giving it another check?

derberg:
@sebastian-palma it works fine, just please move those tests to `test` directory, it will be fine, the purpose of this dir is to keep tests and now we just had there a spec that we used for manual testing but we will anyway remove it in near future as we want to work across all the templates with [this](https://www.asyncapi.com/blog/event-streaming-an-additional-architectural-style-to-suplement-api-design/) dummy example to not "abuse" the streetlights example anymore.

could you copy over in your PR this file https://github.com/asyncapi/generator/blob/master/.github/workflows/pull-request-testing.yml so we run tests with each PR? you can leave the lint part and add a script like `"lint": "echo \"No linter specified yet\""` as we should add eslint soon too

sebastian-palma:
Everything okay @derberg. I added the workflow file and moved the `tests` files to the `test`. I excluded the folders test/output and test/spec since Ada was running them as tests.

Thanks for the help once again ;)

derberg:
@all-contributors please add @sebastian-palma for test

PR #48, opened 2020-07-17
feat: Support asyncapi documents with circular references

derberg:
**Description**

- support asyncapi documents with circular references

testing with:
```
asyncapi: '2.0.0'
info:
  title: Example
  version: 0.1.1
channels:
  recursive:
    subscribe:
      message:
        payload:
          $ref: '#/components/schemas/Recursive'
components:
  schemas:
    Recursive:
      type: object
      properties:
        children:
          type: array
          items:
            $ref: '#/components/schemas/Recursive'
        something:
          type: object
          properties:
            maryna:
              $ref: '#/components/schemas/RecursiveNext'
    RecursiveNext:
      type: object
      properties:
        children:
          type: object
          properties:
            maniek:
              $ref: '#/components/schemas/RecursiveNext'
            ulala:
              type: string
```

final view
html

**Related issue(s)**
Fixes https://github.com/asyncapi/parser-js/issues/83


package.json
```diff
@@ -29,7 +29,7 @@
     "access": "public"
   },
   "dependencies": {
-    "@asyncapi/generator-filters": "^1.0.0"
+    "@asyncapi/generator-filters": "https://github.com/derberg/generator-filters/#handleCircular"
```
fmvilas, line 32: Do not forget about this one.

partials/schema-prop.html
```diff
@@ -107,7 +107,9 @@
       <div class="children {% if odd %}bg-grey-lightest{% else %}bg-grey-lighter{% endif %} py-4 rounded">
         {% if prop.properties() %}
           {% for pName, p in prop.properties() %}
-            {{ schemaProp(p, pName, odd=(not odd), required=(prop.required() | includes(pName))) }}
+            {% if p.isCircular() === false %}
```
fmvilas, line 110: What does it render in case it's circular?

derberg: so far nothing, first we need sampler to support circular refs

fmvilas: But sampler is for examples, right? Would be great to have an indicator saying "Circular" or something. And maybe a link to the part of the page where the circular schema is defined.

derberg: you are right, sorry, was answering without thinking

have a look how it looks after last commit

Screenshot 2020-08-12 at 12 26 51


testing on 
```
asyncapi: '2.0.0'
info:
  title: Example
  version: 0.1.1
channels:
  recursive:
    subscribe:
      message:
        payload:
          $ref: '#/components/schemas/Recursive'
components:
  schemas:
    Recursive:
      type: object
      properties:
        children:
          type: array
          items:
            $ref: '#/components/schemas/Recursive'
        something:
          type: object
          properties:
            maryna:
              $ref: '#/components/schemas/RecursiveNext'
    RecursiveNext:
      type: object
      properties:
        children:
          type: object
          properties:
            maniek:
              $ref: '#/components/schemas/RecursiveNext'
            ulala:
              type: string
```

fmvilas: One super UX feature would be to make `[Circular]` a link that points to where the initial `children` is defined. In this case, it's clear but in large objects might not be that easy to find.

derberg: this is the latest view
html

by link you mean like hyperlink? in this example where would it point?

fmvilas (approved):
🚀

PR #52, opened 2020-08-17
chore: Add twitter publishing to release workflow

derberg:
**Description**

- Add twitter publishing to release workflow


derberg:
Merging without review as it is already approved for the parser and we want to use it everywhere the same way

Issue #69, opened 2020-09-18
Extract deprecation notice to separate macro


derberg:
**Context**

[This](https://github.com/asyncapi/html-template/blob/master/partials/operations/by_tags.html#L34-L50) from `by_tags` and [this](https://github.com/asyncapi/html-template/blob/master/partials/operations/by_tags_no_root.html#L34-L50) from `by_tags_no_root` is exactly the same.

**Description**
1. Next to above mentioned files create a new file, [Nunjucks Macro](https://mozilla.github.io/nunjucks/templating.html#macro)
2. Use this macro in above mentioned files

Example macro -> https://github.com/asyncapi/html-template/blob/master/partials/tags.html
Example macro usage -> https://github.com/asyncapi/html-template/blob/master/partials/message.html#L2 and https://github.com/asyncapi/html-template/blob/master/partials/message.html#L20
Dev instructions -> https://github.com/asyncapi/html-template/tree/master#development﻿

derberg:
video about the issue https://youtu.be/Iqs_2BiNEEo?t=814

HashTalmiz:
Thanks for the video explanation. I'm gonna give this one a try.

derberg:
@HashTalmiz I'm happy that you found the video useful. I'll jump into PR review now

derberg:
fixed with https://github.com/asyncapi/html-template/pull/81

PR #81, opened 2020-10-01
fix: Introduce new macros to keep reusable tag navigation items

HashTalmiz:
This PR refers to Issue #69

**Description**

- Added deprecated.html as a macro
- imported it in by_tags.html and by_tags_no_root.html and used it appropriately

I am not 100% sure if this works. Highly appreciate if the maintainers can look into the changes and let me know if there's a mistake.

### DIFF
--- a/partials/deprecated.html
+++ b/partials/deprecated.html
@@ -1,0 +1,23 @@
+{% macro deprecated(channel, channelName) %}
+
+<a class="js-menu-item flex break-words no-underline text-grey-darkest mt-8 sm:mt-8 md:mt-3"
+    href="#operation-subscribe-{{ channelName }}">
+    {% if channel.deprecated %}
+    <span title="Deprecated"></span>
+    {% endif %}
+    <span class="bg-green-dark font-bold no-underline text-white uppercase p-1 mr-2 rounded"
+        style="height: 21px;font-size: 11px;" title="Subscribe">
+        Sub
+    </span>
+    {% if channel.subscribe().summary() %}
+    <span style="padding-top: 2px;">
+        {{ channel.subscribe().summary() }}
+    </span>
+    {% else %}
+    <div style="display:inline-block;">
+        {{ slicedString(channelName, 'padding-top: 2px;') }}
+    </div>
+    {% endif %}
+</a>
+
+{% endmacro %}
--- a/partials/operations/tag_nav_item.html
+++ b/partials/operations/tag_nav_item.html
@@ -2,9 +2,6 @@
 
 <a class="js-menu-item flex break-words no-underline text-grey-darkest mt-8 sm:mt-8 md:mt-3"
     href="#operation-subscribe-{{ channelName }}">
-    {% if channel.deprecated %}
-    <span title="Deprecated"></span>
-    {% endif %}
     <span class="bg-green-dark font-bold no-underline text-white uppercase p-1 mr-2 rounded"
         style="height: 21px;font-size: 11px;" title="Subscribe">
         Sub
--- a/partials/operations/tag_nav_item.html
+++ b/partials/operations/tag_nav_item.html
@@ -1,4 +1,4 @@
-{% macro deprecated(channel, channelName) %}
+{% macro tagNavItem(channel, channelName) %}
 
 <a class="js-menu-item flex break-words no-underline text-grey-darkest mt-8 sm:mt-8 md:mt-3"
     href="#operation-subscribe-{{ channelName }}">
--- a/partials/operations/by_tags.html
+++ b/partials/operations/by_tags.html
@@ -1,3 +1,5 @@
+{% from "../deprecated.html" import deprecated %}
+
 {% if asyncapi.hasTags() %}
   {% for categorytags in asyncapi.tags() %}
     <div class="mt-4 {% if open %}is-open{% endif %}">
@@ -31,23 +33,7 @@
           {% endif %}
 
           {% if channel.subscribe() and channel.subscribe().hasTags() and channel.subscribe() | containTags(categorytags) %}
-            <a class="js-menu-item flex break-words no-underline text-grey-darkest mt-8 sm:mt-8 md:mt-3" href="#operation-subscribe-{{ channelName }}">
-              {% if channel.deprecated %}
-                <span title="Deprecated"></span>
-              {% endif %}
-              <span class="bg-green-dark font-bold no-underline text-white uppercase p-1 mr-2 rounded" style="height: 21px;font-size: 11px;" title="Subscribe">
-                Sub
-              </span>
-              {% if channel.subscribe().summary() %}
-                <span style="padding-top: 2px;">
-                  {{ channel.subscribe().summary() }}
-                </span>
-              {% else %}
-                <div style="display:inline-block;">
-                  {{ slicedString(channelName, 'padding-top: 2px;') }}
-                </div>
-              {% endif %}
-            </a>
+            {{deprecated(channel, channelName) }}
           {% endif %}
         {% endfor %}
       </div>
--- a/partials/operations/by_tags.html
+++ b/partials/operations/by_tags.html
@@ -1,4 +1,5 @@
-{% from "./tag_nav_item.html" import tagNavItem %}
+{% from "./tag_nav_item_sub.html" import tagNavItemSub %}
+{% from "./tag_nav_item_pub.html" import tagNavItemPub %}
 
 {% if asyncapi.hasTags() %}
   {% for categorytags in asyncapi.tags() %}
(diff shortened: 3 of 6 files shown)


partials/deprecated.html
```diff
@@ -0,0 +1,23 @@
+{% macro deprecated(channel, channelName) %}
+
+<a class="js-menu-item flex break-words no-underline text-grey-darkest mt-8 sm:mt-8 md:mt-3"
+    href="#operation-subscribe-{{ channelName }}">
+    {% if channel.deprecated %}
+    <span title="Deprecated"></span>
+    {% endif %}
```
derberg, line 7: sorry but I just noticed this is some old stuff. AsyncAPI doesn't support `deprecated` property on a channel level and there is no way `channel.deprecated` returns something else than `undefined`.

please remove this part.

next steps would be to
1. rename this file to something like `tag_nav_item.html` and put it in the same dir as files like `by_tags.html `
2. your macro is created perfectly, without issues. Now just extend it a bit. Look, now, when you remove this `deprecated` part [this](https://github.com/asyncapi/html-template/blob/master/partials/operations/by_tags.html#L14-L53) seems to be duplicated twice in both files that you already modified. I think this `children` div  could entirely be moved into this macro. Correct me if I'm wrong

derberg (changes requested):

partials/deprecated.html
HashTalmiz: Yeah you're right. Will add the entire div in the macro.

HashTalmiz: There's a slight difference though in both the divs, [here](https://github.com/asyncapi/html-template/blob/11eb2b093275ac660dadec09ae9c519ae6d62b79/partials/operations/by_tags.html#L1-L2) and [here](https://github.com/asyncapi/html-template/blob/11eb2b093275ac660dadec09ae9c519ae6d62b79/partials/operations/by_tags_no_root.html#L1). The real difference in the macro would just be [here](https://github.com/asyncapi/html-template/blob/11eb2b093275ac660dadec09ae9c519ae6d62b79/partials/operations/by_tags_no_root.html#L16).
I think the only difference would be to give `categorytags` as an argument to the macro for `by_tags.html` and `tag` for `by_tags_no_root.html` Is that correct?

HashTalmiz:
Left a few comments there for you to confirm with

partials/deprecated.html
derberg: I think we probably talk about different things here. I mean that this

```
<div class="children">
        {% for channelName, channel in asyncapi.channels() %}
           {% if channel.publish() and channel.publish().hasTags() and channel.publish() | containTags(categorytags) %}
            <a class="js-menu-item flex break-words no-underline text-grey-darkest mt-8 sm:mt-8 md:mt-3" href="#operation-publish-{{ channelName }}">
              <span class="bg-blue-dark font-bold h-6 no-underline text-white uppercase p-1 mr-2 rounded" style="height: 21px;font-size: 11px;" title="Publish">
                Pub
              </span>
              {% if channel.publish().summary() %}
                <span style="padding-top: 2px;">
                  {{ channel.publish().summary() }}
                </span>
              {% else %}
                <div style="display:inline-block;">
                  {{ slicedString(channelName, 'padding-top: 2px;') }}
                </div>
              {% endif %}
            </a>
          {% endif %}

          {% if channel.subscribe() and channel.subscribe().hasTags() and channel.subscribe() | containTags(categorytags) %}
            <a class="js-menu-item flex break-words no-underline text-grey-darkest mt-8 sm:mt-8 md:mt-3" href="#operation-subscribe-{{ channelName }}">
              {% if channel.deprecated %}
                <span title="Deprecated"></span>
              {% endif %}
              <span class="bg-green-dark font-bold no-underline text-white uppercase p-1 mr-2 rounded" style="height: 21px;font-size: 11px;" title="Subscribe">
                Sub
              </span>
              {% if channel.subscribe().summary() %}
                <span style="padding-top: 2px;">
                  {{ channel.subscribe().summary() }}
                </span>
              {% else %}
                <div style="display:inline-block;">
                  {{ slicedString(channelName, 'padding-top: 2px;') }}
                </div>
              {% endif %}
            </a>
          {% endif %}
        {% endfor %}
      </div>
```

looks like it is used in `by_tags.html` 2 times, and also 2 times in `by_tags_no_root.html`. And it could be moved to a single macro

derberg (changes requested):
@HashTalmiz I added 2 comments, have a look

HashTalmiz:
@derberg Oh now I see what's duplicated. But I don't think I can make a single macro for all 4 cases (including both files) because the top `children` div is different from the bottom `children` div in both files:

For instance: In `by_tags.html`
Top Div Line 16:
```
{% if channel.publish() and channel.publish().hasTags() and channel.publish() | containTags(categorytags) %}
```
And the corresponding statement in the bottom div Line 74:
```
{% if channel.hasPublish() and (not channel.publish().hasTags() or not channel.publish() | containTags(asyncapi.tags())) %}
```
Same thing with these two lines in both divs Line 33 and 91
```
{% if channel.subscribe() and channel.subscribe().hasTags() and channel.subscribe() | containTags(categorytags) %}
{% if channel.hasSubscribe() and (not channel.subscribe().hasTags() or not channel.subscribe() | containTags(asyncapi.tags())) %}
```
I propose 2 solutions:
### **Option 1:**
I think one way to solve this is to make 2 Macros, one for the top/upper `children` div and one for the bottom and use both of them in the 2 pages `by_tags.html` and `by_tags_no_root.html`
## **OR** 
### **Option 2:**
Keep the current macro as it is (i.e., tagNavItem) and make 2 new macros `Sub` and `Pub` to replace [this](https://github.com/asyncapi/html-template/blob/11eb2b093275ac660dadec09ae9c519ae6d62b79/partials/operations/by_tags.html#L17-L30) part and [this](https://github.com/asyncapi/html-template/blob/11eb2b093275ac660dadec09ae9c519ae6d62b79/partials/operations/by_tags.html#L92-L105) part in both pages. (If yes, Please give me better names for these macros lol I'm bad at naming rn)

partials/operations/tag_nav_item.html
```diff
@@ -0,0 +1,20 @@
+{% macro deprecated(channel, channelName) %}
```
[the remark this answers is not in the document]
HashTalmiz: Whoops, forgot to change that, will do it and update the branch to the latest Merge of master in the next commit

derberg:
@HashTalmiz I like option 2. Just a comment that `https://github.com/asyncapi/html-template/pull/81/files#diff-45217e1d33c855e99da7c52cba3e8584` is already [this](https://github.com/asyncapi/html-template/blob/11eb2b093275ac660dadec09ae9c519ae6d62b79/partials/operations/by_tags.html#L92-L105) that you mentioned in your comment, so you can rename `tagNavItem` to `tagNavItemSub` and create another one `tagNavItemPub` from [this](https://github.com/asyncapi/html-template/blob/11eb2b093275ac660dadec09ae9c519ae6d62b79/partials/operations/by_tags.html#L17-L30) so in general there would be 2 macros

HashTalmiz:
Oh yeah forgot that I made that macro, was looking at the master without the changes

derberg:
@HashTalmiz it doesn't work. You need to import this Macro in those 2 new macros `{% from "./sliced-string.html" import slicedString %}`. The files that you refactored didn't need it as they were not macros but includes

I tested with `ag https://raw.githubusercontent.com/asyncapi/generator/master/test/docs/dummy.yml https://github.com/HashTalmiz/html-template/#addMacros -o output -p sidebarOrganization=byTags --force-write`

HashTalmiz:
I did add the changes, and tested my local copy with `ag https://raw.githubusercontent.com/asyncapi/generator/master/test/docs/dummy.yml ./ -o output -p sidebarOrganization=byTags --force-write` It doesn't seem to be working, I get the same error as before.
```
 Error: Unable to call `slicedString`, which is undefined or falsey
```
I thought maybe its `../sliced-string.html` in the import statement, but that doesn't seem to be working either. Am I using the wrong command to test locally?

HashTalmiz:
Well, after making a commit, It is working with `ag https://raw.githubusercontent.com/asyncapi/generator/master/test/docs/dummy.yml https://github.com/HashTalmiz/html-template/#addMacros -o output -p sidebarOrganization=byTags --force-write` 
Maybe it was the wrong command

derberg:
@HashTalmiz sometimes on local you need to enforce reinstallation with `--install` and maybe this was the reason that locally you could not see any improvement. You command was correct.

Anyway, all good. Works as expected. Well done!

derberg (approved):

derberg:
I added `fix` prefix to get patch release triggered as at the end it is not just refactoring as unneeded part about deprecation was removed

PR #82, opened 2020-10-02
feat: Add pdf parameter to generate additional PDF docs

### LINKED ISSUE #72: Add parameter to generate pdf from the HTML file
#### Reason/Context

People still use PDFs to share information, we should support users to easily get PDF out from the HTML. Such functionality is already in [this](https://github.com/derberg/template-for-generator-templates) boilerplate template, so it just needs to be copied.

#### Description

- add `puppeteer` to project dependencies
- add new parameter to template configuration -> https://github.com/derberg/template-for-generator-templates/blob/master/package.json#L62-L65
- creaete new hook responsible for generation, take it from here https://github.com/derberg/template-for-generator-templates/blob/master/hooks/generateExtraFormats.js but just remove pdf and svg generation (these lines https://github.com/derberg/template-for-generator-templates/blob/master/hooks/generateExtraFormats.js#L30-L42)
- update readme

HashTalmiz:
**Description**
- Installed `puppeteer`
- Added hook to generate PDF
- Updated `package.json` to add `pdf` parameter
- Updated Readme

Resolves #72

### DIFF
--- a/hooks/generatePdf.js
+++ b/hooks/generatePdf.js
@@ -10,7 +10,7 @@ async function generatePdf(generator) {
     const parameters = generator.templateParams;
     
     //all actions of this hook depend on parameters passed by the user, if non are provided we should just stop the hook
-    if (!parameters || parameters.pdf!== true) return;
+    if (!parameters || parameters.pdf !== 'true') return;
 
     const browser = await puppeteer.launch();
     const page = await browser.newPage();
--- a/hooks/generatePdf.js
+++ b/hooks/generatePdf.js
@@ -1,0 +1,30 @@
+const puppeteer = require('puppeteer');
+const path = require('path');
+
+module.exports = {
+    'generate:after': generatePdf
+};
+
+async function generatePdf(generator) {
+    const targetDir = generator.targetDir;
+    const parameters = generator.templateParams;
+    //all actions of this hook depend on parameters passed by the user, if non are provided we should just stop the hook
+    if (!parameters) return;
+
+    const browser = await puppeteer.launch();
+    //mermaid has this strange behaviour that it generates the id of the diagram using current date which makes it totally not possible to test
+    //we override date.now() here to make sure always the same number is returned, which is also fine for produciton as only one diagram will be on index.html
+    browser.on('targetchanged', async target => {
+        const targetPage = await target.page();
+        const client = await targetPage.target().createCDPSession();
+        await client.send('Runtime.evaluate', {
+            expression: 'Date.now = function() { return 0; }'
+        });
+    });
+    const page = await browser.newPage();
+    await page.goto(`file://${path.join(targetDir, 'index.html')}`);
+
+    if (parameters.pdf === 'true') await page.pdf({ format: 'A4', path: `${targetDir}/index.pdf` });
+
+    browser.close();
+}
--- a/hooks/generatePdf.js
+++ b/hooks/generatePdf.js
@@ -19,7 +19,7 @@ async function generatePdf(generator) {
 
     await page.$$eval('div.payload-examples > div.js-prop', links => links.forEach(link => link.click()));
     await page.$$eval('div.headers-examples > div.js-prop', links => links.forEach(link => link.click()));
-    await page.$$eval('div.message-examples > div.js-prop', links => links.forEach(link => link.click()));
+    await page.$$eval('div.all-messages div.js-prop', links => links.forEach(link => link.click()));
     
 
     await page.pdf({ format: 'A4', path: `${targetDir}/index.pdf`, printBackground: true });
--- a/hooks/generatePdf.js
+++ b/hooks/generatePdf.js
@@ -8,23 +8,16 @@ module.exports = {
 async function generatePdf(generator) {
     const targetDir = generator.targetDir;
     const parameters = generator.templateParams;
+    
     //all actions of this hook depend on parameters passed by the user, if non are provided we should just stop the hook
-    if (!parameters) return;
+    if (!parameters || parameters.pdf!== true) return;
 
     const browser = await puppeteer.launch();
-    //mermaid has this strange behaviour that it generates the id of the diagram using current date which makes it totally not possible to test
-    //we override date.now() here to make sure always the same number is returned, which is also fine for produciton as only one diagram will be on index.html
-    browser.on('targetchanged', async target => {
-        const targetPage = await target.page();
-        const client = await targetPage.target().createCDPSession();
-        await client.send('Runtime.evaluate', {
-            expression: 'Date.now = function() { return 0; }'
-        });
-    });
     const page = await browser.newPage();
+    
     await page.goto(`file://${path.join(targetDir, 'index.html')}`);
 
-    if (parameters.pdf === 'true') await page.pdf({ format: 'A4', path: `${targetDir}/index.pdf` });
+    await page.pdf({ format: 'A4', path: `${targetDir}/index.pdf` });
 
     browser.close();
 }
--- a/hooks/generatePdf.js
+++ b/hooks/generatePdf.js
@@ -8,13 +8,23 @@ module.exports = {
 async function generatePdf(generator) {
     const targetDir = generator.targetDir;
     const parameters = generator.templateParams;
-    if (!parameters || parameters.pdf === undefined) return;
+    //all actions of this hook depend on parameters passed by the user, if non are provided we should just stop the hook
+    if (!parameters) return;
 
+    const browser = await puppeteer.launch();
+    //mermaid has this strange behaviour that it generates the id of the diagram using current date which makes it totally not possible to test
+    //we override date.now() here to make sure always the same number is returned, which is also fine for produciton as only one diagram will be on index.html
+    browser.on('targetchanged', async target => {
+        const targetPage = await target.page();
+        const client = await targetPage.target().createCDPSession();
+        await client.send('Runtime.evaluate', {
+            expression: 'Date.now = function() { return 0; }'
+        });
+    });
     const page = await browser.newPage();
     await page.goto(`file://${path.join(targetDir, 'index.html')}`);
 
-    if (parameters.pdf === 'true') 
-        await page.pdf({ format: 'A4', path: `${targetDir}/index.pdf` });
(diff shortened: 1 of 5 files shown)


hooks/generatePdf.js
```diff
@@ -0,0 +1,30 @@
...
+    const browser = await puppeteer.launch();
+    //mermaid has this strange behaviour that it generates the id of the diagram using current date which makes it totally not possible to test
+    //we override date.now() here to make sure always the same number is returned, which is also fine for produciton as only one diagram will be on index.html
+    browser.on('targetchanged', async target => {
+        const targetPage = await target.page();
+        const client = await targetPage.target().createCDPSession();
+        await client.send('Runtime.evaluate', {
+            expression: 'Date.now = function() { return 0; }'
+        });
+    });
```
derberg, line 23: as you can see in the comment, this is a solution for the hook in the template where I use mermaid to generate diagrams. Since mermaid is not used here, no need to have this code here either

hooks/generatePdf.js
```diff
@@ -0,0 +1,30 @@
...
+
+module.exports = {
+    'generate:after': generatePdf
+};
+
+async function generatePdf(generator) {
+    const targetDir = generator.targetDir;
+    const parameters = generator.templateParams;
+    //all actions of this hook depend on parameters passed by the user, if non are provided we should just stop the hook
+    if (!parameters) return;
```
derberg, line 12: refactor this `if` and remove `if` from line 27.
in my template I had it done this way as the hook supported 3 different parameters. In case of this hook, it is just this parameter. So basically here in line 12 you should return not only if there are no parameters but also if `parameters.pdf` is not provided

derberg (changes requested):
@HashTalmiz thanks a lot! I added 2 comments to refactor the code a bit

hooks/generatePdf.js
HashTalmiz: Got it

hooks/generatePdf.js
HashTalmiz: Oh yeah that makes sense, will do the changes

HashTalmiz:
That should be all the changes, let me know if anything else needs a change

hooks/generatePdf.js
```diff
@@ -0,0 +1,20 @@
...
+const path = require('path');
+
+module.exports = {
+    'generate:after': generatePdf
+};
+
+async function generatePdf(generator) {
+    const targetDir = generator.targetDir;
+    const parameters = generator.templateParams;
+    if (!parameters || parameters.pdf === undefined) return;
```
derberg, line 11: you removed by accident `const browser = await puppeteer.launch();`


I also think it would be better to replace `parameters.pdf === undefined` with `parameters.pdf !== 'true'` and remove line 16. The generator always gets string, so with the current implementation, so if someone would pass `-p pdf=false` then puppeteer would unnecessarily start a browser.

You can test your work easily on local like this:
- just install generator `npm i -g @asyncapi/generator`
- then run in from the root of your fork like `ag https://bit.ly/asyncapi ./ -o test/output -p pdf=true` or even `ag test/spec/asyncapi.yml ./ -o test/output -p pdf=true` (so you can play with test file from test/spec/asyncapi.yml in case you want to see how the behaviour of template changes)

derberg:
@HashTalmiz have a look at my comment, something is broken there. I also gave you instructions on how to manually run it locally

hooks/generatePdf.js
HashTalmiz: @derberg Thanks for the instructions, In hindsight, I should have run it locally before even making a PR, I do apologise. I've run it locally with the changes you've asked. Although with the `pdf=true` in the command u mentioned, I don't see the difference in the `test/output` file. Is there something wrong or am I looking at the wrong place?

derberg: no worries. You do not get pdf because in your code you expect boolean `true`, while as I wrote, CLI passes always string, so `true` is actually a string `'true'`. I made a code suggestion.

I'm more affraid with the quality of the generated PDF. It has all the data collapsed by default, so if you get a payload with object, you cannot see data of the object. Would make sense to use puppeteer here and before pdf generation, perform a click on collapse/expand chevron across the entire document to get all those expanded. You know what I mean?

HashTalmiz: So what you're trying to say is, to make a preview modal in the html, to preview the pdf before it's generated?

derberg: no, let me explain more in details:

this is how the message under channel is shown, in HTML and then in PDF
Screenshot 2020-10-07 at 09 15 15

so it is good, very good, as payload and headers are expanded. Problem is that in HTML they are not expanded in both, example and messages section:
Screenshot 2020-10-07 at 09 15 06
Screenshot 2020-10-07 at 09 14 55

For HTML it is not a problem as user can click and expand, but yeah, not in case of PDF. 
I suggest we use the power of puppeteer here. It has a rich API to interact with a browser, many people use it to write UI tests (I did in the past too). 
So before you invoke `await page.pdf({ format: 'A4', path: `${targetDir}/index.pdf` });` you can interact with the page and perform `clicks` on those chevrons to expand those `payload` and `headers` sections.

something like this should work to expand all payload examples 
```
await page.$$eval('div.payload-examples > div.js-prop', links => links.forEach(link => link.click()))
```
For headers you would have to eval on `div.header-examples`, I think 😄  and not sure about `Messages` section, I think that maybe [here](https://github.com/asyncapi/html-template/blob/master/partials/messages.html#L8) we need to add some custom class name that could be use in selectors.

with UI operations like this remember that we need to make sure that the page is fully loaded before we execute clicks so modification like this would be needed 
```
await page.goto(`file://${path.join(targetDir, 'index.html')}`, { waitUntil: 'networkidle0' });
```

also add an option to generate background in pdf, should look like 
```
 await page.pdf({ format: 'A4', path: `${targetDir}/index.pdf`, printBackground: true });
```

HashTalmiz: Ah I see what you mean now. 
I did confirm that `hearders-examples` expands the `Headers`, but for the `Message`, adding a `message-example` and making a similar `page.$$eval` doesn't expand the `payload` and `header` chevrons. I've tried placing the class elsewhere but it still didn't work.

hooks/generatePdf.js
```diff
@@ -0,0 +1,28 @@
...
+    if (!parameters || parameters.pdf !== 'true') return;
+
+    const browser = await puppeteer.launch();
+    const page = await browser.newPage();
+
+    await page.goto(`file://${path.join(targetDir, 'index.html')}`, { waitUntil: 'networkidle0' });
+
+    await page.$$eval('div.payload-examples > div.js-prop', links => links.forEach(link => link.click()));
+    await page.$$eval('div.headers-examples > div.js-prop', links => links.forEach(link => link.click()));
+    await page.$$eval('div.message-examples > div.js-prop', links => links.forEach(link => link.click()));
```
derberg, line 22:
```suggestion
    await page.$$eval('div.all-messages div.js-prop', links => links.forEach(link => link.click()));
```

`>` in the selector is used to point to the element that is a direct child so in the examples section, it worked. In the case of messages, `div.js-prop` is a nested child, so just `>` must be removed and  selector will pick all divs, no matter how deep they are

partials/messages.html
```diff
@@ -5,7 +5,7 @@ <h2 class="ml-8 mt-4">Messages</h2>
 
 <div class="responsive-container">
   <div class="center-block p-8">
-    <div class="pb-8">
+    <div class="message-examples pb-8">
```
derberg, line 8:
```suggestion
    <div class="all-messages pb-8">
```

these are not examples but all messages from all the channels

derberg (changes requested):
I made code changes suggestions so you know what was wrong selector for messages section

derberg:
@HashTalmiz you will have to also solve a few merge conflicts because in the morning I merged another contribution with additional new parameter

hooks/generatePdf.js
HashTalmiz: Oh! I misunderstood the meaning of `>` Thanks for the clarification.

HashTalmiz:
What did I GIT? : Since this PR is a branch of my local master, all I did was update local repo's master with upstream (that is this remote repo) and then on my PR branch, I did a `git pull upstream master --rebase` and then once the merge conflicts were resolved, `git push --force-with-lease`. I hope this is the safest way to do it, let me know if it's not. What better practices should I follow for future PRs relating to git?

HashTalmiz:
In hindsight, I think I should have merged with my local repo's master and made my master branch as my PR instead of a side branch. But what made me apprehensive was that I had other branches relating to this repo with other issues so I didn't want to mess up things lol.

derberg:
@HashTalmiz I just tried it locally with `ag https://bit.ly/asyncapi https://github.com/HashTalmiz/html-template/#pdfGeneration -o output -p pdf=true` and it works like a charm. Well done mate!

Regarding GIT. From my point of view the best workflow is:
- when you create a fork, first make sure fork's `master` branch on your local, is linked with `master` branch of the upstream repo. This means that when you do `git pull` on local in `master` branch, you are actually pulling from upstream `master`. I recorded video instruction for this (have a look at [this](https://youtu.be/BsC5tu4M1rw?t=385) particular part of the video)
- every commit should be separated from each other if a different thing is being introduced. In other words, if there are 2 things to do in the PR, then make to separate commits, don't mix. I'm a smart ass here though, I tend to forget about this rule very often. Look at [this](https://github.com/asyncapi/asyncapi-node/pull/20) beautiful PR, 4 things to do = 4 separete commits with good description
- when you have a merge conflict, or simply someone asks you to update our PR with the latest master -> first of all, do not rebase. I always do `git merge master` and then just push the merge commit. Why? because reviewer can clearly see what was changed with a given merge commit, rebase will be invisible

So yeah, for me these are 3 main rules I always try to follow

derberg (approved):

derberg:
@all-contributors please add @HashTalmiz for code

HashTalmiz:
Thank you @derberg for the valuable advice, will make sure to follow them in future PRs!

derberg:
@HashTalmiz no worries, anytime

Issue #56, opened 2020-08-20
Contact email link is rendered wrong


rv0:
#### Describe the bug
The email link in the generated document is wrong.

#### How to Reproduce

Add contact info to the info section

```   
info:
  ...
  contact:
    email: support@asyncapi.org
```   

And render.. The email "link" will be created without "mailto:" prefix, making it a (failing) http link.

#### Expected behavior

The link should be a _mailto:_ link

#### Additional info

The same issue also happens on the playground website.

derberg:
Thanks @rv0 

To anyone who will pick this one up:
- this is where the fix needs to be provided -> https://github.com/asyncapi/html-template/blob/master/partials/introduction.html#L26-L29

derberg:
fixed with #84

PR #83, opened 2020-10-04
Fixed Mailto Link and Empty Payload.

### LINKED ISSUE #55: Support for empty payload
@rv0 commented on [Wed Aug 19 2020](https://github.com/asyncapi/generator/issues/400)

#### Description

"payload" is not mandatory, therefore it can be empty or just left out completely.
However when you do this, the rendering will fail with an unclear error (at least with html-template)

#### How to Reproduce
Remove payload from a message in the streetlights example.

#### Expected behavior
Render empty payload.

---

@jonaslagoni commented on [Wed Aug 19 2020](https://github.com/asyncapi/generator/issues/400#issuecomment-676503738)

@rv0 can you provide the error? This sounds like an issue in the template and not the generator.

---

@derberg commented on [Thu Aug 20 2020](https://github.com/asyncapi/generator/issues/400#issuecomment-677116649)

@jonaslagoni yeap, this is a template related, moving to `html-template` as I know the issue was noticed in the playground

HashTalmiz:
2 Birds, One stone!
**Description**
- Added `mailto` to the `href` for mail support as mentioned in issue #56 
- Wrapped `msg.payload()` within an if statement as in issue #55 

Resolves #55 AND #56

### DIFF
--- a/partials/deprecated.html
+++ b/partials/deprecated.html
@@ -1,0 +1,23 @@
+{% macro deprecated(channel, channelName) %}
+
+<a class="js-menu-item flex break-words no-underline text-grey-darkest mt-8 sm:mt-8 md:mt-3"
+    href="#operation-subscribe-{{ channelName }}">
+    {% if channel.deprecated %}
+    <span title="Deprecated"></span>
+    {% endif %}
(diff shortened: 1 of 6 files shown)


HashTalmiz:
Some of the file changes are from my other PRs which are awaiting review so let those be done first

PR #86, opened 2020-10-07
feat: Add parameter to change output HTML file name

### LINKED ISSUE #73: Add support to generate filename different than index.html
#### Reason/Context

There is a use case where you want to have documentation for all your services stored in one folder, which means you cannot stick to `index.html` file as you will override those files. you need to be able to set a different name for the output file. Sample use case is described here https://github.com/asyncapi/generator/issues/331

#### Description

such functionality is already added to the Markdown template, so it is all about copying it over and modifying just a bit:
- add new parameter to html template config, just like https://github.com/asyncapi/markdown-template/blob/master/package.json#L68-L72
- add a hook that after generation renames `index.html` to whatever name you provide with `outFilename` parameter, just like this https://github.com/asyncapi/markdown-template/blob/master/hooks/output-filename.js
- add new parameter to the readme

bufutda:
**Description**
Added an optional parameter `outFilename` (default: `index.html`) to rename the output file.

**Related issue(s)**
Resolves #73


derberg (approved):
perfect job, thanks!

derberg:
@all-contributors please add @bufutda for code

PR #94, opened 2020-10-08
fix: Default to Authorization header when auth type is http

### LINKED ISSUE #28: Header name should default to Authorization when auth type is http
@fmvilas commented on [Wed Jan 22 2020](https://github.com/asyncapi/generator/issues/201)

**Applies to:** HTML and Markdown templates.

---

Given this security definition:

```yaml
components:
  securitySchemes:
    bearerAuth:
      type: http
      scheme: bearer
      bearerFormat: JWT
      description: The JWT for authorizing against the backend. Needs to be obtained through the authentication route.
```

It's not possible to define the name of the header but the default value "Authorization" should be shown. At least until we fix [this](https://github.com/asyncapi/asyncapi/issues/307). Currently, we got this:

image

Thanks @luca-moser for reporting the bug 🙌 

---

@github-actions[bot] commented on [Wed May 13 2020](https://github.com/asyncapi/generator/issues/201#issuecomment-627664589)

This issue has been automatically marked as stale because it has not had recent activity :sleeping:
It will be closed in 30 days if no further activity occurs. To unstale this issue, add a comment with detailed explanation. 
Thank you for your contributions :heart:

anbreaker:
**Description**

- Default to Authorization when auth type is http.

**Related issue(s)**
Fixes #28

Captura de pantalla de 2020-10-08 13-14-58

### DIFF
--- a/partials/servers.html
+++ b/partials/servers.html
@@ -106,7 +106,7 @@
 
               {% if def.bearerFormat() %}
                 <span class="font-bold no-underline bg-blue-light text-white text-xs uppercase rounded"
-                  style="height: 20px;font-size: 11px;padding: 3px;">Header: {{'Authorization' if def.type() === 'http' else def.name()}}
+                  style="height: 20px;font-size: 11px;padding: 3px;">Header: Authorization
                 </span>
               {% endif %}
 
--- a/partials/servers.html
+++ b/partials/servers.html
@@ -106,7 +106,8 @@
 
               {% if def.bearerFormat() %}
                 <span class="font-bold no-underline bg-blue-light text-white text-xs uppercase rounded"
-                  style="height: 20px;font-size: 11px;padding: 3px;">Header: {{def.name()}}</span>
+                  style="height: 20px;font-size: 11px;padding: 3px;">Header: {{'Authorization' if def.type() === 'http' else def.name()}}
+                </span>
               {% endif %}
 
               {% if def.openIdConnectUrl() %}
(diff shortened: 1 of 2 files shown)


partials/servers.html
```diff
@@ -106,7 +106,8 @@ <h5 class="text-sm text-grey mt-1">Security:</h5>
 
               {% if def.bearerFormat() %}
                 <span class="font-bold no-underline bg-blue-light text-white text-xs uppercase rounded"
-                  style="height: 20px;font-size: 11px;padding: 3px;">Header: {{def.name()}}</span>
+                  style="height: 20px;font-size: 11px;padding: 3px;">Header: {{'Authorization' if def.type() === 'http' else def.name()}}
```
derberg, line 109: when I look at [the spec](https://github.com/asyncapi/asyncapi/blob/master/versions/2.0.0/asyncapi.md#fixed-fields-21) to under the meaning of `bearerFormat` and `name` I say it is enough to fix it like this

```suggestion
                  style="height: 20px;font-size: 11px;padding: 3px;">Header: Authorization
```

Pretty hardcoded but look:
- `bearerFormat` is only ment for `http` type so this `def.type() === 'http'` is not needed as no other security object will have this field
- `name` is only for `httpApiKey` type so `http` type will never have it so this `else def.name()` is never going to happen

Yeah I just realized it during the review... sorry for the misleading comment under issue. Does it make sense?

derberg (changes requested):
@anbreaker yey, first step! now every next one will be much easier. Have a look at my comment to your solution

anbreaker:
Ok ok thanks, I understood the new way of thought.

fmvilas (approved):

derberg (approved):
LGTM! Now waiting for another PR 😄

PR #99, opened 2020-10-14
fix: scalar types should not have a chevron icon indicating you can expand it

### LINKED ISSUE #50: [BUG] Scalar types are shown as expandable when they aren't
#### Describe the bug
Scalar types like `String`, `Int`, `Boolean`, etc. are shown as "expandable" when there's nothing to "expand".

#### How to Reproduce
Check out [playground.asyncapi.io]() with Streetlights example.

#### Expected behavior
They shouldn't show the "expand" arrow button.

gabrielclaudino:
**Description**

The scalar types are coming with an empty object in `patternProperties()` thus triggering the `isExpandable` filter. I changed the condition to verify if the object returned by `patternProperties()` is empty or not. It may be valid to also check why they are coming with it in the first place.

**Related issue**
Resolves #50

### DIFF
--- a/filters/all.js
+++ b/filters/all.js
@@ -14,7 +14,7 @@ function isExpandable(obj) {
     (fun(obj.additionalProperties) && obj.additionalProperties()) ||
     (fun(obj.extensions) && obj.extensions() &&
       Object.keys(obj.extensions()).filter(e => !e.startsWith("x-parser-")).length) ||
-    (fun(obj.patternProperties) && obj.patternProperties())
+    (fun(obj.patternProperties) && Object.keys(obj.patternProperties()).length)
   ) return true;
 
   return false;
--- a/test/all.is_expandable.test.js
+++ b/test/all.is_expandable.test.js
@@ -4,6 +4,7 @@ const {isExpandable} = require("../filters/all");
 test.beforeEach(t => {
   t.context.doc = {};
   t.context.types = [{ type: "string" }, { type: "number" }];
+  t.context.doc.patternProperties = () => ({});
 });
 
 test("isExpandable when the object type is object", t => {


derberg (changes requested):
This one was a tricky beast 😄 
Please also adjust the tests. As you can see they were passing no matter what change you did, even though we have a test case that should fail -> https://github.com/asyncapi/html-template/blob/master/test/all.is_expandable.test.js#L64-L66

Schema model always has `patternProperties` function, and it always returns `{}` unless there is something. This is why your fix works now, but to have a reliable test it should have  `t.context.doc.patternProperties = () => ({});` in the beforeEach function in between those lines https://github.com/asyncapi/html-template/blob/master/test/all.is_expandable.test.js#L5-L6

gabrielclaudino:
Thank you for the quick review. I changed the tests :)

derberg (approved):
Thanks a lot @gabrielclaudino 

LGTM

I only changed the title to something clearer for people looking later in the release notes

gabrielclaudino:
Okay!

derberg:
@all-contributors please add @gabrielclaudino for code

derberg:
@gabrielclaudino I would appreciate if you could share how did you learn about this project. Was it our blog post about Hacktoberfest, or Discord info. What resources were helpful, which one not. Would love to know your experience so we know how to make it better for next Hacktoberfest and where to advertise our involvement.

gabrielclaudino:
Sorry for the delay, I hadn't seen your question. I first heard about your project in the hacktoberfest Discord. I was browsing the project-sharing channel, and your post caught my attention. The issues spreadsheet help me a lot to pick a issue that suits my skills (more js related).

derberg:
@gabrielclaudino thanks for taking the time to answer 🙏

PR #104, opened 2020-10-16
docs: clean up example spec and dev guide

### LINKED ISSUE #74: Cleanup example spec and development guide
#### Reason/Context

We should use the same example spec across all templates as much as possible to test them always against the same example. There is such a dummy file in the Generator repo -> https://github.com/asyncapi/generator/blob/master/test/docs/dummy.yml

#### Description
- remove https://github.com/asyncapi/html-template/blob/master/test/spec/asyncapi.yml
- adjust dev guide https://github.com/asyncapi/html-template#development
- adjust script https://github.com/asyncapi/html-template/blob/master/package.json#L26 to use asyncapi file from remote location `ag https://raw.githubusercontent.com/asyncapi/generator/master/test/docs/dummy.yml ./ -o test/output --force-write --watch-template`

bszwarc:
**Description**

- Removed `https://github.com/asyncapi/html-template/blob/master/test/spec/asyncapi.yml` file
- Pointed to the right file in the dev guide
- Adjusted `package.json` file
- Fixed minor language slips

Question: I would suggest adding a colon to the sentences with commands in the `Development` section to separate commands from text. It can also be organized in other way, like having commands below text. 

**Related issue(s)**
Resolves #74


README.md
```diff
@@ -27,10 +27,10 @@ npm install -g @asyncapi/generator
 
 ## Development
 
-1. Make sure you have the latest generator installed `npm install -g @asyncapi/generator`.
-1. Modify the template or it's helper functions. Adjust `test/spec/asyncapi.yml` to have more features if needed.
-1. Generate output with watcher enables `npm run develop`.
-1. Open HTML in your browser `open ./test/output/index.html`.
+1. Make sure you have the latest generator installed: `npm install -g @asyncapi/generator`.
+1. Modify the template or its helper functions. Adjust [`dummy.yml`](https://raw.githubusercontent.com/asyncapi/generator/master/test/docs/dummy.yml) to add more features if needed.
```
derberg, line 31: please change the link here and in the other file to `https://raw.githubusercontent.com/asyncapi/generator/v1.0.0-rc.12/test/docs/dummy.yml` as we finally have it under a stable tag. Pointing directly to master is risky as it can change in master to a form that can cause errors in the template

derberg, line 31: Please add a sentence for contributors to let them know that if they had to modify a dummy file to include more content (to see the feature they worked on in action) they should open a PR to change this fine https://github.com/asyncapi/generator/blob/master/test/docs/dummy.yml in generated repository

derberg (changes requested):
@bszwarc awesome stuff 🙇 just 2 tiny requests

README.md
bszwarc: Agree - done :)

bszwarc: Added :)

README.md
```diff
@@ -27,10 +27,10 @@ npm install -g @asyncapi/generator
 
 ## Development
 
-1. Make sure you have the latest generator installed `npm install -g @asyncapi/generator`.
-1. Modify the template or it's helper functions. Adjust `test/spec/asyncapi.yml` to have more features if needed.
-1. Generate output with watcher enables `npm run develop`.
-1. Open HTML in your browser `open ./test/output/index.html`.
+1. Make sure you have the latest generator installed: `npm install -g @asyncapi/generator`.
+1. Modify the template or its helper functions. Open a PR in the `asyncapi/generator` repository to add more content to the [`dummy.yml`](https://raw.githubusercontent.com/asyncapi/generator/v1.0.0-rc.12/test/docs/dummy.yml) file.
```
derberg, line 31: better add this link https://github.com/asyncapi/generator/blob/master/test/docs/dummy.yml. Not all people know how to get from a raw link to a repo link

also `asyncapi/generator` would be better as [asyncapi/generator](https://github.com/asyncapi/generator)

> Open a PR in the `asyncapi/generator` repository to add more content to the [`dummy.yml`](https://raw.githubusercontent.com/asyncapi/generator/v1.0.0-rc.12/test/docs/dummy.yml) file.

don't you think this sentence requires a condition? like it is not that you always need to open a PR, you should do it only if the dummy file was not full-featured for you and during development you had to modify it

bszwarc: For the link, sure - I was a bit confused maybe with the previous comment about changing the link in this file and the other one to the stable tag version. But yes, in general you are right, let me handle this. 
I also added the link to the repo  and got rid of the code formatting for now (unless you have specific guidelines on that).
Question: To me, the entire step 2 in the `Development` section implies the modification because someone wants to extend it with the new content / feature and run it.  To do so, the user needs to open a PR. 
In the file is fine for the user and does not require any modifications he would simply use the template as is 🤔  We can modify the sentence nonetheless to `If you want to add new content.... open a PR` or something along these lines, to make it more conditional.

derberg: dummy file is just an example of AsyncAPI document, one of many that could be consumed with this `html-template`. We call it dummy as it should not be used as a reference for AsyncAPI users to learn AsyncAPI specification with it. Its purpose is to have as complex file as possible with as many different AsyncAPI components as possible. This way it is much easier to work on features in templates, ans you do not have to invents sample data file but just take the dummy one that covers majority of AsyncAPI spec features.

so basically the dev guide is about only html-template changes. Thing is that not all features might be present in the dummy file, so to develop html-template you need to add some stuff to dummy file locally (as sample data) to have easier life. But then you should contribute those extra things to repo where dummy file is (what can be missing from the dummy file? example https://github.com/asyncapi/generator/issues/429)

so "possible" dummy file PR is not part of this guide, it is a conditional change that you might need to do. Maybe it should be added as a quote after step 2?

bszwarc: Got it now, thanks!

derberg (approved):
LGTM!!!

Thanks you 🙏

derberg:
@all-contributors please add @bszwarc for docs

PR #109, opened 2020-10-20
fix: display parameter description first instead of its schema description

### LINKED ISSUE #26: Parameter description does not render in resulting output with html template
@maxs commented on [Thu Mar 12 2020](https://github.com/asyncapi/generator/issues/249)

In the folowing example `description` field won't be rendered in resulting output because it is not taken from the parameter `toy` but the `schema`.  This means, to have `description`  rendered, we would have to put `description` under `schema` block which is not intended behavior.  

```yml
components:
  parameters:
    toy:
      description: Description of the toy
      schema:
        type: string
```

---

@github-actions[bot] commented on [Thu Mar 12 2020](https://github.com/asyncapi/generator/issues/249#issuecomment-598331101)

Welcome to AsyncAPI. Thanks a lot for reporting your first issue.

Keep in mind there are also other channels you can use to interact with AsyncAPI community. For more details check out [this issue](https://github.com/asyncapi/asyncapi/issues/115).

---

@derberg commented on [Thu Mar 12 2020](https://github.com/asyncapi/generator/issues/249#issuecomment-598332646)

Some additional hint:
When I quickly look into the code of HTML template, the `schemaProp` macro https://github.com/asyncapi/generator/blob/master/templates/html/.partials/operation.html#L45 gets only the schema object. So the description that is used in the template `https://github.com/asyncapi/generator/blob/master/templates/html/.partials/schema-prop.html#L76` is not taken from the `parameter` but the `schema`. 

---

@github-actions[bot] commented on [Wed May 13 2020](https://github.com/asyncapi/generator/issues/249#issuecomment-627664554)

This issue has been automatically marked as stale because it has not had recent activity :sleeping:
It will be closed in 30 days if no further activity occurs. To unstale this issue, add a comment with detailed explanation. 
Thank you for your contributions :heart:

anbreaker:
**Description**
Add, condition to watch Alt description if exist this variable.

**Related issue(s)**
Fixes #26


partials/schema-prop.html
```diff
@@ -1,7 +1,8 @@
 {% from "./type.html" import type %}
 {% from "./schema-item.html" import schemaItem %}
+{% from "./description.html" import getDescription %}
```
derberg, line 3: @anbreaker this is the part I didn't cover in the video, so I do not do all 😄 

as you can see, file `description.html` doesn't exist, you need to create it, and put in there a macro called `getDescription` that will return such a div element `<div class="text-sm markdown">{{ description | markdown2html | safe }}</div>` with proper value of course. More about macros -> https://mozilla.github.io/nunjucks/templating.html#macro

good luck, and remember to test to make sure it works before you let me know to have another look

derberg (changes requested):

partials/schema-prop.html
```diff
@@ -80,9 +81,8 @@
           </div>
         </div>
 
-        {% if prop.description() %}
-          <div class="text-sm markdown">{{ prop.description() | markdown2html | safe }}</div>
-        {% endif %}
+        {{ getDescription(altDescription or prop.description()) }}
```
derberg, line 84: above this line add something like `{{ altDescription | log}}` and tell me what you see in logs in termina

derberg (approved):
Good work!

Issue #112, opened 2020-10-27
Crash when generating pdf


flokasper:
#### Describe the bug

I am trying to generate pdf from the asyncapi.json using docker, rendering the html works perfectly but I have a crash when generating the pdf:
```
npm http fetch GET 200 https://registry.npmjs.org/@asyncapi%2fhtml-template 894ms
npm http fetch GET 200 https://registry.npmjs.org/@asyncapi/html-template/-/html-template-0.15.1.tgz 77ms
npm http fetch GET 304 https://registry.npmjs.org/rimraf 71ms (from cache)
npm http fetch GET 200 https://registry.npmjs.org/rimraf/-/rimraf-3.0.2.tgz 38ms
npm http fetch GET 200 https://registry.npmjs.org/puppeteer 115ms
npm http fetch GET 200 https://registry.npmjs.org/puppeteer/-/puppeteer-5.4.0.tgz 75ms
npm http fetch GET 200 https://registry.npmjs.org/@asyncapi%2fgenerator-filters 712ms
npm http fetch GET 200 https://registry.npmjs.org/@asyncapi/generator-filters/-/generator-filters-1.1.0.tgz 56ms
npm http fetch GET 200 https://registry.npmjs.org/openapi-sampler 37ms
npm http fetch GET 200 https://registry.npmjs.org/markdown-it 44ms
npm http fetch GET 200 https://registry.npmjs.org/markdown-it/-/markdown-it-10.0.0.tgz 29ms
npm http fetch GET 200 https://registry.npmjs.org/openapi-sampler/-/openapi-sampler-1.0.0-beta.17.tgz 435ms
npm http fetch GET 304 https://registry.npmjs.org/entities 36ms (from cache)
npm http fetch GET 200 https://registry.npmjs.org/entities/-/entities-2.0.3.tgz 28ms
npm http fetch GET 200 https://registry.npmjs.org/json-pointer 39ms
npm http fetch GET 200 https://registry.npmjs.org/json-pointer/-/json-pointer-0.6.1.tgz 30ms
npm http fetch GET 200 https://registry.npmjs.org/foreach 33ms
npm http fetch GET 200 https://registry.npmjs.org/foreach/-/foreach-2.0.5.tgz 33ms
npm http fetch GET 200 https://registry.npmjs.org/https-proxy-agent 43ms
npm http fetch GET 200 https://registry.npmjs.org/proxy-from-env 74ms
[...]

PR #113, opened 2020-10-28
fix: add args to puppeteer to make pdf generation work in docker

derberg:
**Description**

- add args to puppeteer to make pdf generation work in docker

**Related issue(s)**
Fixes https://github.com/asyncapi/html-template/issues/112


jonaslagoni (approved):
👍

PR #114, opened 2020-10-29
fix: mark required props in object in array

derberg:
**Description**

- mark required props in object in array

**Related issue(s)**
See also https://github.com/asyncapi/playground/issues/43


jonaslagoni (approved):
👍

PR #117, opened 2020-11-09
fix: array with array of items throws error and only object is supported

derberg:
**Description**

- array with an array of items throws an error and only object is supported

I think I introduced the bug when I was adding support for circular refs. Good thing is though that we now regularly extend our dummy yaml file with more cases that need to be covered in tools -> https://github.com/asyncapi/generator/pull/443

**Related issue(s)**
Fixes https://github.com/asyncapi/generator/issues/441

### DIFF
--- a/partials/schema-prop.html
+++ b/partials/schema-prop.html
@@ -152,7 +152,8 @@
         {% endif %}
 
         {% if prop.type() === 'array' %}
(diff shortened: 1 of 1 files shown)


jonaslagoni (approved):
Tried with the given use-case in the issue, works👍

PR #125, opened 2021-01-12
ci: remove release workflow and testing to use the ones that are global

derberg:
**Description**

This PR will be merged and right after it, the global replacement:
- https://github.com/asyncapi/.github/pull/18
- https://github.com/asyncapi/.github/pull/21

**Related issue(s)**
See also https://github.com/asyncapi/asyncapi/issues/423


.github/workflows/pull-request-testing.yml
```diff
@@ -7,18 +7,21 @@ on:
 jobs:
```
derberg, line 7: we could turn it into global workflow called `pull-request-testing-if-nodejs.yml` and by adding [this](https://github.com/asyncapi/.github/pull/21/files#diff-e8ca070981ccdc3f369851c960bec6882143a0c7afe78acd7b05e0c2b0068d72R18-R20) it would work only on nodejs projects. What do you think?


so name would be `PR testing if Nodejs`

advantage, switching of node 14 to 16 would be done in just one file again

We could even do the same with release workflow in general 🤔

magicmatatjahu: > We could even do the same with release workflow in general

Good idea, but remember that in `react-component` we have a totally different workflow for release due to 3 packages: normal component, web-component + playground

magicmatatjahu: Ok, I see in another PR's comments, that you know about this problem 😄

fmvilas (approved):

Issue #126, opened 2021-01-13
Minify and purge CSS files


Theiaz:
Hello all! Today I've found this nice project and want to give some ideas.

#### Reason/Context

Having a look at `tailwind.min.css` it seems like there has been no pruning. File size is nealry 400KB and I think there are much unused styles. Without pruning one big advantge of TailwindCSS is missed ([similar case](
https://v1.tailwindcss.com/docs/installation#using-tailwind-via-cdn)). I think the file size would be reduced significantly. For further information have a look at [TailwindCSS documentation](https://v1.tailwindcss.com/docs/controlling-file-size).

Additionally `main.css` could also be minified to reduce file size.

derberg:
@Theiaz hey, you mean you'd like to have a kind of post-processing of the resulting generation output? or is there something that needs to be improved in the HTML files to enable any custom post-processing anyone can add?

Theiaz:
@derberg Hello

I think its necessary to introduce post-processing with the generated ouput. HTML files/partials looks fine to me. 

There is no need to include a separate `tailwind.min.css`. Instead you can setup the tailwind dependency via npm, add a `tailwind.config.js` and adjust `main.css` (if I'm right). Have a look at [Installation Guide](https://v1.tailwindcss.com/docs/installation).  Maybe I can try to help with a pull request if I got time.

Take care I've linked the old version (v1) of TailwindCSS to prevent breaking changes.

derberg:
PR would be welcome. Triggering of post-processing could be done in an automated way as we do with PDF generation -> https://github.com/asyncapi/html-template/blob/master/hooks/generatePdf.js

Theiaz:
Seems like your `tailwind.min.css` is using very old styles and is not more compatible with newest TailwindCSS versions. For example there is no more class `.bg-blue-dark`. I do not know where the current styles are from, however it seems pretty similar to [this](https://github.com/tailwindlabs/tailwindcss-plugin-examples/blob/master/dist/styles.css).

I will investigate what to do if the newest Tailwind versions needs to be used. For sure, the CSS classes of the partials needs to be adjusted.

PR #127, opened 2021-01-15
feat: use tailwind v2 and decrease css size with purge and minify

Theiaz:
Regarding to [this](https://github.com/asyncapi/html-template/issues/126) issue TailwindCSS needs to be a dependecy to purge unused styles.

As I mentioned, the old stylesheet has used an deprecated version of tailwind styles. Therefore, some tailwind classes were outdated and needs to be adjusted (so far I've found colors and `pin-`). For exchanging the colors there were two apporaches.

1. Use the old styles with copy & paste it into the `main.css` stylesheet
1. Use the new color classes tailwind provides

I've decided to go for the second option, cause its more maintainable and less effort. Otherwise I would have to search all the classes with its regarding tailwind prefixes (like `focus:`, `active:`, ..) and paste them all into `main.css`. Drawback of this approach is that the old color hexcodes may differ a lilttle bit. However, I tried to keep the visual difference as minimal as possible.

Color mapping has been changed the following way. Each basic color  is mapped to its `-500` value. For example `text-grey` -> `text-gray-500`. The other color gradations are mapped as following:

`color-lightest` -> `color-200`
`color-lighter` -> `color-300`
`color-light` -> `color-400`
`color` -> `color-500`
`color-dark` -> `color-600`
`color-darker` -> `color-700`
`color-darkest` -> `color-800`

Currently its using version `1.x.x` of tailwind instead of the new `2.x.x` for easier transition.

As a result the size of the CSS file is reduced drastically (only 26kb).  

Resolves: #126


derberg:
@Theiaz great stuff man 🥳 

Can you please have a look on the sonar cloud errors?
Also, what did you have to change in the dummy example so now you point to `https://raw.githubusercontent.com/asyncapi/generator/73dd4d18f51c20a962118dabc40f0c8068ccec20/test/docs/dummy.yml`?

You updated in readme the dev flow, what about user flow? I mean who is responsible for running `npm run buildStyles` and when it should be done?

Theiaz:
The sonarcloud errors for `@tailwind` directive is a false positive. However, I can not mark it, maybe you could resolve it? The two code smells will be fixed soon,

Running the old dummy example leads to an error, cause the version of this file has been outdated. New file is pointing to current dummy file of master.

User flow: You're right, this needs to be done. In my opinion it would be best, if `npm run buildStyles` is triggered after each successful build. Could this be a job for CI/CD pipeline?

derberg:
@Theiaz marked those errors as resolved. I will jump into review this week

derberg:
@Theiaz hey, can you solve merge conflicts, please 🙏🏼

Theiaz:
To solve those conflicts I need to know if ` "generate:assets"` is the desired CI/CD step for building the styles? Or whats the purpose of this command?

derberg:
@Theiaz `generate:assets` are for everything additional that has to be generated for projects, API docs/ toc in readme, examples, bundles, type script types - that kind of things

derberg:
it is a generic script to run them all, if needed, otherwise just needs to be there with echo that nothing is there

README.md
```diff
@@ -32,6 +32,7 @@ npm install -g @asyncapi/generator
 
     >**NOTE:** If you have to modify the [`dummy.yml`](https://github.com/asyncapi/generator/blob/master/test/docs/dummy.yml) file to develop your features, open a PR with the changes in the [asyncapi/generator](https://github.com/asyncapi/generator) repository.
 
+1. Adjust styling and generate `tailwind.min.css` with `npm run buildStyles`
```
derberg, line 35: we need update here

package.json
```diff
@@ -22,17 +22,20 @@
     "lint": "echo \"No linter specified yet\"",
     "test": "ava",
     "release": "semantic-release",
-    "develop": "ag https://raw.githubusercontent.com/asyncapi/generator/v1.0.0-rc.12/test/docs/dummy.yml ./ -o test/output --force-write --watch-template",
-    "generate:assets": "echo 'No additional assets need to be generated at the moment'",
+    "develop": "ag https://raw.githubusercontent.com/asyncapi/generator/73dd4d18f51c20a962118dabc40f0c8068ccec20/test/docs/dummy.yml ./ -o test/output --force-write --watch-template",
```
derberg, line 25: why this specific version from this exact commit?

package.json
```diff
@@ -44,6 +47,9 @@
     "conventional-changelog-conventionalcommits": "^4.2.3",
     "semantic-release": "^17.0.4"
   },
+  "peerDependencies": {
+    "postcss": "^8.2.4"
```
derberg, line 51: just out of curiosity, why peerDependency? and not dev?

derberg (commented):
few initial comments

Theiaz:
I've resolved your reviewed issues. Sorry for the mess in my latest commit, there are some changes (resolved issues) which do belong to other commits. 

I've migrated tailwind to v2 to be ready for future. However, to avoid breaking changes I need to manually wire in the color and font-sizes of v1, cause v2 had some changes. You can see this additional config in `tailwind.config.js`. The UI still looks like with v1.

package.json
derberg: @Theiaz you didn't answer my curiosity 😄

derberg (commented):
Just checked with `ag https://bit.ly/asyncapi-dummy https://github.com/Theiaz/html-template#126-purge-and-minify-css --force-write` and styling is almost perfect. Have a look at the `h1` style, it is slightly different. Also, there is a problem with the position of the chevron responsible for expand/collapse

Screenshot 2021-01-29 at 10 03 10

When you run generation locally it is good to use `--install` flag during development, especially when you generate once using the original template and then right again you generate using your fork (but the template name is still the same)

derberg:
I can see we will go down from 302KB to 22KB 🥳

Theiaz:
Nice Hawkeye! I've found two more minor styling issues which I fixed. Hopefully its mergeable now :)

derberg (approved):
@Theiaz all works perfectly now. Well done mate!

If you share with me your Twitter and LinkedIn details I'd love to mention you on our social accounts, so it is clear who contributed this important feature.

derberg:
@all-contributors please add @Theiaz for code

Theiaz:
> @Theiaz all works perfectly now. Well done mate!
> 
> If you share with me your Twitter and LinkedIn details I'd love to mention you on our social accounts, so it is clear who contributed this important feature.

Thanks!
I've updated my profile: @ju_schaefer

derberg:
@Theiaz fyi
- https://www.linkedin.com/feed/update/urn:li:activity:6763414026811715584
- https://twitter.com/AsyncAPISpec/status/1357648468798504965

PR #141, opened 2021-02-07
fix: include main.css if not params.singleFile

marcortw:
Not a 100% sure, but I guess omitting main.css prohibits some commonmark renderings

### DIFF
--- a/template/index.html
+++ b/template/index.html
@@ -20,6 +20,7 @@
   {% else %}
   <link href="css/tailwind.min.css" rel="stylesheet">
   <link rel="stylesheet" href="css/atom-one-dark.min.css" />
+  <link rel="stylesheet" href="css/main.css" />
   {% endif %}
 </head>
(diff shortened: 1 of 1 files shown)


derberg:
@marcortw main.css is part of tailwind.min.css because of https://github.com/asyncapi/html-template/blob/master/package.json#L28. @Theiaz is looking into it

marcortw:
@derberg , ok if I close this PR because this file is anyway part of tailwind and not going to be changed here?

derberg:
@marcortw up to you. It also might be that your fix is the only possible. I mean that `main.css` was there before introduction of `postcss`. This css file is not large so from my perspective, your fix is a kind of last resort if @Theiaz is not able to find a better solution

derberg:
@marcortw now we can close it as https://github.com/asyncapi/html-template/pull/146 fixed the issue

PR #144, opened 2021-02-09
docs: documentation on how to skip chromium installation

stavalfi:
according to a suggestion in https://github.com/asyncapi/generator/issues/480#issuecomment-776026732

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ npm install -g @asyncapi/generator
 |outFilename|The filename of the output file.|No|*Any*|`asyncapi.html`|
 |pdf|Generates output HTML as PDF|No|`true,false`|`false`|
 
-If you only generate an html website, set the environment variable `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` to `true` and the generator will skip downloading chromium. The generation time will dramatically decrease when not using npm as a package-manager.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,6 +25,8 @@ npm install -g @asyncapi/generator
 |outFilename|The filename of the output file.|No|*Any*|`asyncapi.html`|
 |pdf|Generates output HTML as PDF|No|`true,false`|`false`|
 
+If you only generate an html website, set the environment variable `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` to `true` and the generator will skip downloading chromium. The generation time will dramatically decrease when not using npm as a package-manager.
```
derberg, line 28:
```suggestion
If you only generate an html website, set the environment variable `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` to `true` and the generator will skip downloading chromium.
```

I removed `The generation time will dramatically decrease when not using npm as a package-manager.` as this has nothing to do with the flat, html-template is still fetched with npm, it is just that the chromium installation is skipped.

derberg:
@stavalfi thanks a lot for such a fast PR. Just one comment added

derberg (approved):

PR #146, opened 2021-02-12
fix: rendering of list items in markdown

### LINKED ISSUE #142: Rendering of CommonMark does not produce expected output
#### Describe the bug
When I create an output of the streetlights example, I don't get anymore nicely rendered output as it used to be. I thought it had something to do with a missing main.css file (see [this PR](https://github.com/asyncapi/html-template/pull/141)), but fixing this locally didn't do the trick.

#### How to Reproduce
Create a new repo and `npm install @asyncapi/generator@1.1.7`
Create a minimal example like this:

```
const path = require("path");
const Generator = require("@asyncapi/generator");
const generator = new Generator("@asyncapi/html-template",path.resolve(__dirname, "example"),{ forceWrite: true });

(async () => {
  try {
    await generator.generateFromURL(
      "https://raw.githubusercontent.com/asyncapi/asyncapi/2.0.0/examples/2.0.0/streetlights.yml"
    );
    console.log("Done!");
  } catch (e) {
    console.error(e);
  }
})();
```
This produces output like this:
image

#### Expected behavior
I used to get output like this (this worked until @asyncapi/html-template@0.15.4 but not anymore afterwards):
image

Notice the bullet items which come from this CommonMark formatted section in the spec:

```
  description: |
    The Smartylighting Streetlights API allows you to remotely manage the city lights.

    ### Check out its awesome features:

    * Turn a specific streetlight on/off 🌃
    * Dim a specific streetlight 😎
    * Receive real-time information about environmental lighting conditions 📈
```

Theiaz:
**Description**
Due to the migration from tailwind v0.x to v2.x changed its default styling for lists ([see here](https://www.tailwindcss.cn/docs/upgrading-to-v1/#8-remove-any-usage-of-list-reset)). This does not affect the UI in general (like the `<ul>`in the sidebar), because list items were already rendered without bullet points. However, source from markup code should be displayed with bullet points and be intended.

**Related issue(s)**
Resolves #142

### DIFF
--- a/partials/servers.html
+++ b/partials/servers.html
@@ -49,7 +49,7 @@
 
       {% if server.security() %}
         <h5 class="text-sm text-gray-500 mt-1">Security:</h5>
-        <ul class="list-reset">
+        <ul>
           {% for sec in server.security() %}
             <li>
               {% set def = asyncapi.components().securityScheme(sec.json() | keys | head) %}
--- a/partials/servers.html
+++ b/partials/servers.html
@@ -111,7 +111,7 @@
               {% endif %}
 
               {% if def.openIdConnectUrl() %}
-                <div class="px-4 py-2 ml-4 mb-3 border border-gray-400 bg-gray-200 rounded">
+                <div class="px-4 py-2 ml-4 mb-3 border border-gray-400 bg-gray-100 rounded">
                   <div>
                     <span class="text-xs font-bold text-gray-600 mt-1 mr-1 uppercase">Connect URL:</span>
                     <a class="text-gray-600 text-xs font-normal" href="{{def.openIdConnectUrl()}}"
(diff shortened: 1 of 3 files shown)


template/css/tailwind.min.css
```diff
@@ -1176,6 +1176,12 @@ a:hover {
   text-transform: uppercase;
 }
 
+.markdown :is(ol, ul) {
+  -webkit-padding-start: 40px;
+          padding-inline-start: 40px;
+  list-style-type: disc;
+}
+
```
derberg, line 1184: should we have some customizations in `tailwind.min.css`? no other way?

derberg: @Theiaz ok, I know why it is here, because you locally ran `npm run generate:assets` and then you committed changes to tailwind. Just please remove and we release. In another PR would be awesome somehow fix it, so postcss is actually processing stuff into a new file that is not in the repo, that we can add to gitignore

thoughts?

Theiaz: Yeay your're right. `tailwind.min.css` should be generated from ci/cd. I deleted the file.

derberg (approved):
@Theiaz all good now, thanks a lot for the fix man!

PR #149, opened 2021-02-15
fix: postinstall of css artifact

derberg:
**Description**

Atm artifacts are not generated during release and not pushed to npm. This PR fixes it and introduces a separate css file that is considered to be `generated-only`

@chatne thanks for spotting the issue

- introduce `main.min.css` that will always be `.gitignore` and will be a generated artifact.
- add `postinstall` script that always generates fresh `main.min.css`  file
- use only `main.min.css`  in the HTML, no matter the mode (singleFile or not)
- `main.min.css` will not be pushed to npm as it will be anyway always generated after installation + it will never be updated in the repository

@Theiaz @marcortw  what do you think about this change?

**Related issue(s)**
See also https://github.com/asyncapi/html-template/issues/142#issuecomment-778768780


marcortw:
That sounds reasonable to me, thanks @derberg!

package.json
```diff
@@ -22,10 +22,14 @@
...
     "release": "semantic-release",
-    "develop": "ag https://raw.githubusercontent.com/asyncapi/generator/v1.1.5/test/docs/dummy.yml ./ -o test/output --force-write --watch-template",
+    "develop": "npm run postinstall && ag https://raw.githubusercontent.com/asyncapi/generator/v1.1.5/test/docs/dummy.yml ./ -o test/output --force-write --watch-template",
     "//develop:install": "Force ag to install your local html-template source code",
     "develop:install": " npm run develop -- --install",
-    "generate:assets": "cross-env NODE_ENV=production postcss template/css/main.css -o template/css/tailwind.min.css",
+    "generate:atomcss": "cross-env NODE_ENV=production postcss template/css/atom-one-dark.min.css -o template/css/main.min.css",
+    "generate:maincss": "cross-env NODE_ENV=production postcss template/css/custom.css -o template/css/main.min.css",
+    "generate:tailwind.css": "cross-env NODE_ENV=production postcss template/css/tailwind.min.css -o template/css/main.min.css",
+    "postinstall": "rimraf \"template/css/main.min.css\" && npm run generate:atomcss && npm run generate:maincss && npm run generate:tailwind.css",
```
magicmatatjahu, line 31: I have one question. Does postcss merge everything to one `main.min.css`, yes?

derberg: define `everything` 😄 there is no way to merge multiple files into one at once, nothing in the CLI that I've seen 😞

magicmatatjahu: So, `postinstall` script override `main.min.css` 3 times, or what? 😄

derberg: appends/merges

derberg: I mean not a simple append but smart merge 😄

magicmatatjahu (approved):
LGTM

Theiaz:
Sorry for being late.

I had a look on the changes and I want to leave my opinion:

> * introduce main.min.css that will always be .gitignore and will be a generated artifact.

As far as i can see, that changes has already been reverted in #153 
> * add postinstall script that always generates fresh main.min.css file

Merging all css files into one is a good idea. However you do not need to do it this way. There are easier options like [@import](https://developer.mozilla.org/en-US/docs/Web/CSS/@import). Postcss can handle this and its [working fine with tailwind](https://tailwindcss.com/docs/using-with-preprocessors#build-time-imports).  We just needs to use `postcss-import` and adjust our file structure:

We just need two files: `atom-one-dark.css` and `main.css` with the following structure:

```css
@import "./atom-one-dark.css"; // import

@tailwind base;

@tailwind components;

@tailwind utilities;

// our custom css
html,
body { ... }

....
```

I've tested it locally and will open a PR.

>* use only main.min.css in the HTML, no matter the mode (singleFile or not)

Thats fine!

>* main.min.css will not be pushed to npm as it will be anyway always generated after installation + it will never be updated in the repository

What about separation of development files and generated assets / build files? Would it be possible to place our "output" files like `main.min.css` inside a seperate `dist` folder  or do they really needs to be inside `template/? Possibly this can be extended to `.js` and `.html`  files aswell? There would be no need to push `dist` folder anymore and generated assets could easily be dropped.

derberg:
@Theiaz hey man, no worries, thanks for getting back with valuable input and a PR!

`template` dir is a must as the generator only touches files that are there in the template, we can have `template/dist` of course, and at the end `dist` goes to the output

derberg:
generator support hooks into the generation process, we can have a hook that after generation removes not needed files from the output. We could actually move postcss processing to hooks too if this is needed and if possible to use postcss as a library. Here you have an example of the hook https://github.com/asyncapi/html-template/blob/master/hooks/02_renameOutFile.js

PR #159, opened 2021-02-19
fix: use css imports instead of separate postcss processing of each file

### LINKED ISSUE #157: 0.18.0 is missing hljs css
#### Describe the bug

the html is missing hljs styles

#### How to Reproduce

```sh
npx @asyncapi/generator --force-write --output test \
  https://raw.githubusercontent.com/asyncapi/generator/v1.1.5/test/docs/dummy.yml \
  @asyncapi/html-template
```

The HTML is missing syntax highlighting in the examples

image

### 0.17.1 worked

this is a regression from 0.17.1, where syntax highlighting worked

```sh
npx @asyncapi/generator --force-write --output test \
  https://raw.githubusercontent.com/asyncapi/generator/v1.1.5/test/docs/dummy.yml \
  @asyncapi/html-template@0.17.1
```

image

### LINKED ISSUE #158: Use css imports for merging css files
As described [here](https://github.com/asyncapi/html-template/pull/149#issuecomment-782114843) instead of merging css files by npm scripts a better way would be to use `@import`

Theiaz:
**Description**
Instead of merging css files by npm scripts a better way would be to use `@import` inside css files.

I've also unminimized the `atom-one-dark.min.css`, because it should be human readable. Minification should only be done for `main.min.css` files which is the only included css file.  

**Related issue(s)**
fixes #157
fixes #158


derberg:
@Theiaz why `tailwind.min.css ` is removed? is it covered by https://github.com/asyncapi/html-template/pull/159/files#diff-05ed1c7f99e45b485708115502a1e0f28c8547e9a9a29c1b8664f79103cf7873R4 ?

Theiaz:
Separate css file for tailwind styles like `tailwind.min.css` is not necessary. As you can see in `main.min.css` all the needed utility classes of tailwind are generated (like `px-2`, `text-gray-100`, ...). 

This is because of the `@tailwind base/components/utility` directives inside `main.css`. Tailwind scannes the css class declarations inside the html files. Files which needs to be be scanned are declared  inside `tailwind.config.js`. 

```js
module.exports = {
  purge: ["./template/**/*.html", "./partials/**/*.html"], 
  ...
}
```

 Tailwind then automatically wires the necessary style inside `main.css`. The resulting css file has the following structure:

```css
/* atom-dark theme */
/* tailwind utility classes */
/* custom css declared after @tailwind utilities; */
```

derberg:
@Theiaz perfect, thanks for the patience and detailed explanation 🙏🏼

derberg (approved):

PR #181, opened 2021-04-15
feat: support spec extension rendering for messages

derberg:
**Description**

- refactor current extensions rendering support to have it as reusable macro
- for now support it only on a message, other can come when others need it, I needed it for message only
- modified css to make sure dump of long text in pre code is wrapped

Screenshot 2021-04-15 at 10 39 03
Screenshot 2021-04-15 at 10 38 50

### DIFF
--- a/partials/extensions.html
+++ b/partials/extensions.html
@@ -1,0 +1,44 @@
+{% macro extension(obj, odd = false, indent = false) %}
+
+{% set extensions = obj | nonParserExtensions%} 
+{% if extensions.size > 0 %}
+      <p class="{% if indent %}pl-6{% endif %} mb-2 mt-4 text-sm font-bold text-gray-700">
+        Specification Extensions:
+      </p>
+      {% for extensionName, extensionValue in extensions %}
+      <div class="{% if odd %}bg-gray-100{% else %}bg-gray-200{% endif %} {% if indent %}pl-6{% endif %}">
+        {% if extensionValue | isObject %}
+        <div class="{% if open %}is-open{% endif %}">
+          <div class="js-prop cursor-pointer py-2 property">
+            <div class="pr-4" style="margin-top: -2px; min-width: 25%">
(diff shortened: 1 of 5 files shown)


partials/extensions.html
```diff
@@ -0,0 +1,44 @@
+{% macro extension(obj, odd = false, indent = false) %}
+
+{% set extensions = obj | nonParserExtensions%} 
+{% if extensions.size > 0 %}
+      <p class="{% if indent %}pl-6{% endif %} mb-2 mt-4 text-sm font-bold text-gray-700">
+        Specification Extensions:
+      </p>
+      {% for extensionName, extensionValue in extensions %}
+      <div class="{% if odd %}bg-gray-100{% else %}bg-gray-200{% endif %} {% if indent %}pl-6{% endif %}">
```
derberg, line 9: `{% if indent %}pl-6{% endif %}` is the only change I introduced to the code after extracting it from schema prop macro, to support conditional indent of spec extension data that sometimes is needed in the UI

partials/extensions.html
```diff
@@ -0,0 +1,46 @@
+{% macro extension(obj, odd = false, indent = false) %}
+
+{% set extensions = obj | nonParserExtensions%} 
+{% if extensions.size > 0 %}
+    <p class="{% if indent %}pl-6{% endif %} mb-2 mt-4 text-sm font-bold text-gray-700">
+    Specification Extensions:
+    </p>
+
+    {% for extensionName, extensionValue in extensions %}
+        <div class="{% if odd %}bg-gray-100{% else %}bg-gray-200{% endif %} {% if indent %}pl-6{% endif %}">
```
derberg, line 10: `{% if indent %}pl-6{% endif %}` is the only change I introduced to the code after extracting it from schema prop macro, to support conditional indent of spec extension data that sometimes is needed in the UI

magicmatatjahu (approved):

Issue #165, opened 2021-03-18
Circular reference causes a "Template render error" error


several27:
Hey all,

New user here, absolutely loving the asyncapi and the generator. Thanks for all the effort!

Not 100% sure if the below is a bug or us defining the schema incorrectly, but would really appreciate a little bit of help. Just for the context: we're trying to express a schema, similar to how most data processing libraries like spark define their type system (recursive struct with fields). 

#### Describe the bug

Circular reference causes a "Template render error" error

#### How to Reproduce

```
asyncapi: 2.0.0
info:
  title: Circular Reference test
  version: 1.0.0
  description: Smiley face
channels:
  /:
    subscribe:
      message:
        $ref: '#/components/messages/SchemaMessage'
components:
  messages:
    SchemaMessage:
      payload:
        $ref: '#/components/schemas/Schema'
  schemas: 
    Schema:
      type: object
      properties: 
        kind: 
          oneOf:
            - $ref: '#/components/schemas/Schema'
            - type: string
            - enum:
              - boolean
              - string 
```

<img width="1792" alt="Screen Shot 2020-11-22 at 8 57 25 PM" src="https://user-images.githubusercontent.com/3248329/99930285-57705c80-2d05-11eb-97ca-0786dc1b27c9.png">

#### Expected behavior

I believe a similar issue has been solved here: https://github.com/asyncapi/parser-js/issues/83#issuecomment-673921723

It would be amazing if the generator at least didn't crash on the above schema unless it's incorrect. 

Thank you so much!

magicmatatjahu:
Similar issue #67 

It's related to issue in the ParserJS repo - asyncapi/parser-js#293 We need to wait for the PR in the mentioned repo.

PR #286, opened 2021-11-24
feat: support circular references

magicmatatjahu:
**Description**

- Support circular references
- Adding `global.min.css` file with reset global styles (adding fonts etc)
- Rename `styles.min.css` to the `asyncapi.min.css`
- Move `@asyncapi/parser` dep to the dependencies
- Update tests

New version of template will only run with latest generator `>= 1.8.27`

**Related issue(s)**
Blocked by https://github.com/asyncapi/generator/pull/721
Resolves https://github.com/asyncapi/html-template/issues/165


filters/all.js
```diff
@@ -98,7 +62,7 @@ filter.includeFile = includeFile;
  * and annotates that specification is parsed.
  */
 function stringifySpec(asyncapi) {
-  return JSON.stringify(replaceCircular(asyncapi.json()));
+  return asyncapi.constructor.stringify(asyncapi);
```
derberg, line 65: calling itself to stringify itself 🤯

magicmatatjahu:
@derberg I can make it in different way - install `@asyncapi/parser` dependency in template and then use `AsyncAPIDocument.stringify()` function, but here I use hack to read the definition of instance (I mean AsyncAPIDocument class) from `constructor` property and then execute needed function. What do you think? Go with current solution or with dependency?

test/filters/all.test.js
```diff
@@ -66,19 +66,15 @@ describe('Filters', () => {
               type: 'object',
               properties: {
                 foo: { type: 'string', 'x-parser-schema-id': '<anonymous-schema-1>' },
-                bar: {
-                  // x-parser-schema-id and x-parser-circular-props are inferred from dummySchema
-                  // stringifySpec infers all `x-parser` extensions
-                  'x-parser-schema-id': 'dummySchema',
-                  'x-parser-circular-props': ['bar']
-                }
+                bar: '$ref:$.components.schemas.dummySchema',
```
derberg, line 69: what is this strange format, where is it coming from? can you share some light?

derberg:
@magicmatatjahu `but here I use hack to read the definition of instance` - I think the word `hack` already answers what is my preference 😄 . Parser is already used in this project for tests, so just move it to `Dependencies`

magicmatatjahu:
@derberg I cannot answer to you in the thread (I don't know why) so I will write here. I made changes in parser-js to support stringify the spec. Here is a doc for that https://github.com/asyncapi/parser-js#stringify. This "strange" format is a JSON Pointer which points where exactly we have in JSON this same object (reference to it), so we preserver circular references in string. Very similar approach to the `json-schema-ref-parser` package, but it's very small solution (and fast). You probably can ask (like Jonas asked me), why not going with `json-schema-ref-parser`? Two reason:
- this package cannot handle bundling of our spec (it can only handle JSON Schemas)
- this package adds some ~X0kb (pre bundling has 150kb so after bundling it can add 40kb?) to the final bundle, so we will end with bigger rendered template (we need functionality to parse the spec)

Also as I wrote, our solution is very simple, small and fast - you can check revision https://github.com/asyncapi/parser-js/commit/dcee64944343f6fb74566e3905680d133946332f

derberg:
@magicmatatjahu yeah, I looked into docs and noticed info about JSON Pointer, but there is nothing about `$ref:` which is confusing, so it is not a pure JSON Pointer.

magicmatatjahu:
@derberg The `$ref` is a prefix that specifies that the value is a json pointer. This makes it easier to check if a value is a regular string or a json pointer. You're right, I'll add that to the docs, I forgot about it. Thanks! 🦅 👁️

derberg:
@magicmatatjahu give it a thought, `$ref` does not indicate it is JSON Pointer, it indicates it is a reference 😄 
IMHO `$.` is pretty "indicative" enough 😄 For me it is confusing, as you already saw it.

magicmatatjahu:
Okay, I was fast 😅  Right `$ref` means there is a reference, but this prefix is added to tell if it was previously referenced or not (during stringify). If we went without the prefix then someone could write a value in the spec starting with `$.` and we'd have a problem, because how do we check if we're going to find a reference or treat it as just a value? 🤷🏼

magicmatatjahu:
@derberg I created PR for docs https://github.com/asyncapi/parser-js/pull/418 :)

magicmatatjahu:
@derberg 
- moved parser to the dependencies
- updated PR's description
- added global styles for removing default browser styling - we need that after new styling in react-component
- created PR for docs asyncapi/parser-js#418

Could you check again? :)

derberg (approved):

PR #321, opened 2022-02-02
fix: move to synchronouse deletion of unnecessary files

magicmatatjahu:
**Description**

As in title. In the [server-api](https://github.com/asyncapi/server-api) project we have problem when someone pass `singleFile` parameter and due to fact that given files are removed in async mode (without waiting) we can end up with situation when someone will get generated template with these files.


derberg (approved):
:shipit:

PR #254, opened 2021-08-24
feat: add option to override react config.

btimby:
I have updated `README.md`, I am testing this change locally right now. I will follow up with results shortly.

**Description**

- Adds an option for `html-template` to define react config.
- Can be used to disable sidebar.

**Related issue(s)**

https://github.com/asyncapi/html-template/issues/253

PR #363, opened 2022-08-19
feat: add param to override react component config

BTOdell:
README updated. Tests added. Sorry for the extra lines in the diff, my editor automatically formatted the files.

**Description**

- Adds a param to override the React component config prop
- Supersedes PR #254 (which is stale). Applied feedback from that PR by @magicmatatjahu 

**Related issue(s)**
#254


README.md
```diff
@@ -36,14 +36,15 @@ npm install -g @asyncapi/generator
 
 ## Supported parameters
 
-| Name | Description | Required | Default | Allowed values |  Example |
+| Name | Description | Required | Default | Allowed values | Example |
 |---|---|---|---|---|---|
 | sidebarOrganization | Defines how the sidebar should be organized. Set its value to `byTagsNoRoot` to categorize operations by operations tags. Set its value to `byTags` when you have tags on a root level. These tags are used to model tags navigation and need to have the same tags in operations. | No | undefined | `byTags`, `byTagsNoRoot` | `byTagsNoRoot` |
 | baseHref | Sets the base URL for links and forms. | No | `/` | *Any* | `/docs` |
-| version | Override the version of your application provided under `info.version` location in the specification file. | No | Version is taken from the spec file. | *Any* ([See Semver versionning](https://semver.org/)) | `1.0.0` |
+| version | Override the version of your application provided under `info.version` location in the specification file. | No | Version is taken from the spec file. | *Any* ([See Semver versioning](https://semver.org/)) | `1.0.0` |
```
magicmatatjahu, line 43: Thanks! 😅

README.md
```diff
@@ -36,14 +36,15 @@ npm install -g @asyncapi/generator
...
 | singleFile | Set output into one html-file with styles and scripts inside | No | `false` | `true`,`false` | `true` |
 | outFilename | The filename of the output file. | No | `index.html` | *Any* | `asyncapi.html` |
 | pdf | Generates output HTML as PDF | No | `false` | `true,false` | `false` |
+| config | Inline stringified JSON or a path to a JSON file to override default React component config | No | `null` | [JSON config for the React component](https://github.com/asyncapi/asyncapi-react/blob/next/docs/configuration/config-modification.md#definition) | `{"show":{"sidebar":false}}` |
```
magicmatatjahu, line 47: Default value for that parameter should be `{ show: { sidebar: true }, sidebar: { showOperations: 'byDefault' } }`. Also we should mention merging of config works:

```suggestion
| config | Inline stringified JSON or a path to a JSON file to override default React component config. The default config and passed via patameter is merged using the [JSON Merge Patch](https://tools.ietf.org/html/rfc7386) algorithm. | No | `{ "show": { "sidebar": true }, "sidebar": { "showOperations": 'byDefault' } }` | [JSON config for the React component](https://github.com/asyncapi/asyncapi-react/blob/next/docs/configuration/config-modification.md#definition) | `{"show":{"sidebar":false}}` |
```

filters/all.js
```diff
@@ -7,15 +7,52 @@ const { AsyncAPIDocument } = require('@asyncapi/parser');
...
   const config = { show: { sidebar: true }, sidebar: { showOperations: 'byDefault' } };
+  // Apply config override
+  if (params.config) {
+    let configOverride;
+    try {
+      configOverride = JSON.parse(params.config);
+    } catch (err) {
+      // Failed to parse JSON, attempt to read as JSON file
+      configOverride = JSON.parse(fs.readFileSync(params.config, "utf8"));
+    }
```
magicmatatjahu, line 47: From what I can see you will be operating on an absolute path and not a local path. It would be nice to fix this somehow to give the option of two options of local path and absolute path - I'm not sure how this `readFile` function works underneath, if it tries to check the absolute path and then the local path in case of an error. Can you enlighten me? We should first check for absolute path and then check local using `cwd`.

Also it would be nice to make some `console.error` before throwing error that given config isn't valid JSON/file path.

package.json
```diff
@@ -103,6 +103,11 @@
       "pdf": {
         "description": "Set to `true` to get index.pdf generated next to your index.html",
         "default": false
+      },
+      "config": {
+        "description": "Stringified JSON or path to JSON file to override [React component config](https://github.com/asyncapi/asyncapi-react/blob/next/docs/configuration/config-modification.md#definition)",
+        "default": null,
```
magicmatatjahu, line 109: It should be string:

```suggestion
        "default": "",
```

package.json
```diff
@@ -103,6 +103,11 @@
       "pdf": {
         "description": "Set to `true` to get index.pdf generated next to your index.html",
         "default": false
+      },
+      "config": {
+        "description": "Stringified JSON or path to JSON file to override [React component config](https://github.com/asyncapi/asyncapi-react/blob/next/docs/configuration/config-modification.md#definition)",
```
magicmatatjahu, line 108:
```suggestion
        "description": "Stringified JSON or a path to a JSON file to override default React component config. The default config and passed via patameter is merged using the [JSON Merge Patch](https://tools.ietf.org/html/rfc7386) algorithm.",
```

magicmatatjahu (changes requested):
Awesome! We need to improve description of new parameter and I described my problem with changed logic of `prepareConfiguration` function. Overall great, but some things need to be addressed and fixed.

filters/all.js
BTOdell: All relative paths are resolved against the current working directory in the Node.js `fs` module. The code above first tries to parse the config as stringified json, then falls back to assuming the config string is a file path and attempts to read it. I agree that we should add a second try catch around the file loading and report a friendly error message. Should it be a fatal error or just print and continue?

package.json
BTOdell: done

README.md
BTOdell: 😄

README.md
BTOdell: done

package.json
BTOdell: I didn't use the exact same text as "The default config and passed via patameter is merged using..." but let me know if what I came up with is acceptable.

filters/all.js
BTOdell: Updated. Let me know if the second try-catch is sufficient to provide a friendly error message with `console.error`. Also added more comments to explain what the code does.

magicmatatjahu: > All relative paths are resolved against the current working directory in the Node.js fs module. The code above first tries to parse the config as stringified json, then falls back to assuming the config string is a file path and attempts to read it. I agree that we should add a second try catch around the file loading and report a friendly error message. Should it be a fatal error or just print and continue?

Thanks for clarification! Something light in my head regarding the default cwd but I couldn't quite remember.

> Updated. Let me know if the second try-catch is sufficient to provide a friendly error message with console.error. Also added more comments to explain what the code does.

Thanks! `console.error` is sufficient, but I think that after logging we should throw error (`err` or `jsonErr` but `err` will be better) to stop executing rest of the code - generator should stop there and not continue generation.

package.json
magicmatatjahu: My English is not perfect, so thanks a lot! :)

filters/all.js
BTOdell: Done.

BTOdell:
Anything else preventing this from being merged?

magicmatatjahu (approved):
LGTM! @BTOdell Sorry for such a delay!

Issue #367, opened 2022-09-13
Allow setting favicon


JarnoRFB:
#### Reason/Context
To brand the generated docs a little bit, it would be nice to be able to set a favicon.

#### Description
The basic idea is to pass a URL via a parameter and use that as a favicon. A further option would be to get the favicon from a local file.

PR #368, opened 2022-09-13
feat: Allow setting favicon

JarnoRFB:
**Description**
Sets the favicon by default to the asyncapi favicon and allows to override it with a custom URL.

**Related issue(s)**
Resovles #367


derberg (approved):
@JarnoRFB thanks for the contribution!

LGTM 🚀 

@magicmatatjahu anything you wanna add?

package.json
```diff
@@ -104,6 +104,10 @@
         "description": "Set to `true` to get index.pdf generated next to your index.html",
         "default": false
       },
+      "favicon": {
+        "description": "URL of the favicon",
+        "default": "https://www.asyncapi.com/favicon.ico"
```
magicmatatjahu, line 109: Could we also add possibility to pass favicon from local disk? We should check if given path is local/absolute in filesystem and then copy it to the output folder. Do you wanna handle that?

JarnoRFB: I think the option definitely makes sense. I was not sure though, how the copying would exactly work. Also the sinlge file option would have to handled differently, maybe by encoding the file as base64. I can look into that though.

magicmatatjahu: You can copy something to the final output using `generate:after` hook like here https://github.com/asyncapi/html-template/blob/master/hooks/02_renameOutFile.js in mentioned file you can see how we deal with targetDir (output path) and passed parameters (templateParams). As for `singleFile` feature - yeah we should try to convert favicon to the base64 and use `data:image/png;base64,{icon}` format. I can make it for you if you don't have time :)

magicmatatjahu:
@JarnoRFB Ok, I pushed some changes. Now we generate from favicon the `base64` data and apply it to the `<head>`. Now we support local favicons and external ones :) Thanks for contribution! I know that sometimes we don't have a time for Open Source, so don't worry. I had no problem with helping you, and sorry for such a delay!

magicmatatjahu:
@JarnoRFB You can check code, I will wait for Monday and then I will merge it, ok? :)

magicmatatjahu (approved):

JarnoRFB:
@magicmatatjahu Thanks for taking this up! Was unfortunately to busy to continue working on.

magicmatatjahu:
@JarnoRFB No problem man! :) Thanks for contribution! If you will use template with new parameter and if you will find bugs, please let us know! :)

PR #388, opened 2022-11-03
feat: added timeout parameter to pdf generator

### LINKED ISSUE #387: Add timeout parameter to use in pdf generation
#### Reason/Context
Please try answering few of those questions
- Why we need this improvement?
- How will this change help?
- What is the motivation? 
Some times the pdf generation consumes more than the default timeout used in puppeteer that is 30 segs. So this change will allow to generate the PDF's that consumes more time than the default.

#### Description
Please try answering few of those questions
- What changes have to be introduced?
- Will this be a breaking change?
- How could it be implemented/designed?

I will add a templateParam to the package.json file as well as handling for said parameter in the HTML template with the default value of 30000 ms. Add the timeout parameter in Page.pdf with the new parameter value.
I will add documentation for this new parameter.

This will not be a breaking change. No real design is necessary.

llobetrj:
**Description**

- Added timeout parameter and pass to the Page.pdf as parameter

**Related issue(s)**

Resolves #387

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,8 @@ npm install -g @asyncapi/generator
 | version | Override the version of your application provided under `info.version` location in the specification file. | No | Version is taken from the spec file. | *Any* ([See Semver versioning](https://semver.org/)) | `1.0.0` |
 | singleFile | Set output into one html-file with styles and scripts inside | No | `false` | `true`,`false` | `true` |
 | outFilename | The filename of the output file. | No | `index.html` | *Any* | `asyncapi.html` |
-| pdf | Generates output HTML as PDF | No | `false` | `true,
-| timeout | Timeout (in ms) used to generate the PDF | No | 30000 | >=0 | `false` |false` | `false` |
+| pdf | Generates output HTML as PDF | No | `false` | `true`, `false` | `true` |
+| pdfTimeout | Timeout (in ms) used to generate the PDF | No | 30000 | >=0 | 1000 |
 | config | Inline stringified JSON or a path to a JSON file to override default React component config. The config override is merged with the default config using the [JSON Merge Patch](https://tools.ietf.org/html/rfc7386) algorithm. | No | `{ "show": { "sidebar": true }, "sidebar": { "showOperations": "byDefault" } }` | [JSON config for the React component](https://github.com/asyncapi/asyncapi-react/blob/next/docs/configuration/config-modification.md#definition) | `{"show":{"sidebar":false}}` |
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -43,7 +43,8 @@ npm install -g @asyncapi/generator
 | version | Override the version of your application provided under `info.version` location in the specification file. | No | Version is taken from the spec file. | *Any* ([See Semver versioning](https://semver.org/)) | `1.0.0` |
 | singleFile | Set output into one html-file with styles and scripts inside | No | `false` | `true`,`false` | `true` |
 | outFilename | The filename of the output file. | No | `index.html` | *Any* | `asyncapi.html` |
-| pdf | Generates output HTML as PDF | No | `false` | `true,false` | `false` |
+| pdf | Generates output HTML as PDF | No | `false` | `true,
+| timeout | Timeout (in ms) used to generate the PDF | No | 30000 | >=0 | `false` |false` | `false` |
```
magicmatatjahu, line 47: It's not a timeout of template, but for pdf generation :) We should use `pdfTimeout` to clarify what it is. Also you have broken md table here.

```suggestion
| pdf | Generates output HTML as PDF | No | `false` | `true`, `false` | `true` |
| pdfTimeout | Timeout (in ms) used to generate the PDF | No | 30000 | >=0 | 1000 |
```

hooks/99_generatePdf.js
```diff
@@ -31,7 +31,7 @@ module.exports = {
         toClick && typeof button.dispatchEvent === 'function' && button.dispatchEvent(new Event('click', { bubbles: true }));
       }));
     
-      await page.pdf({ format: 'A4', path: `${targetDir}/index.pdf`, printBackground: true });
+      await page.pdf({ format: 'A4', path: `${targetDir}/index.pdf`, printBackground: true, timeout: parseInt(templateParams.timeout,10) });
```
magicmatatjahu, line 34:
```suggestion
      await page.pdf({ format: 'A4', path: `${targetDir}/index.pdf`, printBackground: true, timeout: parseInt(templateParams.pdfTimeout, 10) });
```

package.json
```diff
@@ -104,6 +104,11 @@
         "description": "Set to `true` to get index.pdf generated next to your index.html",
         "default": false
       },
+      "timeout":{
```
magicmatatjahu, line 107:
```suggestion
      "pdfTimeout": {
```

magicmatatjahu (changes requested):
Good, but we should change name of paramater.

Issue #430, opened 2023-06-05
AsyncAPI spec v3 support in HTML-template


jonaslagoni:
### Reason/Context
This Issue is used to track changes needed to support AsyncAPI v3. As a code owner, please edit this list of TODO tasks in order to properly track the progress 🙂 Once this issue is closed it means that v3 is now fully supported in this library.  

### Scope
- [x] Migrate to the new JS parser API
- [ ] https://github.com/asyncapi/asyncapi-react/issues/733

fmvilas:
/progress 90 Just waiting for https://github.com/asyncapi/html-template/pull/448 to get merged.

jonaslagoni:
/progress 80 waiting for parser to be released to update dependencies and then it should work out the box

Issue #185, opened 2021-04-21
Move to JSX renderer


magicmatatjahu:
#### Reason/Context

Currently template is written in Nunjuck. We should move to JSX (aka React) renderer, of course after [unification](https://github.com/asyncapi/html-template/issues/51). Unfornatelly we are blocked by https://github.com/asyncapi/generator/issues/521. After improving compilation of React templates, switching to React template will be very easy.

magicmatatjahu:
Still walid. I will annotate that by `good first issue` label.

ritik307:
@derberg @magicmatatjahu I would like to work on it but can you kindly elaborate on the issue

magicmatatjahu:
@ritik307 Hello! Yeah, no problem, we can discuss. Main improvement here will be change Nunjucks templates to the JSX files (React components). Atm our generator supports two render engines, mentioned Nunjucks and JSX - more info about JSX and Nunjucks is here:
- JSX https://github.com/asyncapi/generator/blob/master/docs/authoring.md#react-1
- Nunjucks https://github.com/asyncapi/generator/blob/master/docs/authoring.md#nunjucks

We have also https://github.com/asyncapi/template-for-generator-templates repo when you can check that same templates written in Nunjucks and JSX to check difference how to achieve that same output:
- Nunjucks https://github.com/asyncapi/template-for-generator-templates/blob/nunjucks/template/index.html
- JSX https://github.com/asyncapi/template-for-generator-templates/blob/master/template/index.js 

We are opened for contributions :)

So some actions points to that issue:
- you should change that file https://github.com/asyncapi/html-template/blob/master/template/index.html to the JSX template
- we should still support all parameters that we have for that template https://github.com/asyncapi/html-template#supported-parameters
- all hooks should also work after rewriting template

It sounds as big change, but you can make it by iteration, please rewriting in first step the mentioned file to JSX and we can discuss about next changes :)

ritik307:
hi @magicmatatjahu correct me if am wrong, initially, I have to convert `index.html` to a JSX template `index.jsx`

magicmatatjahu:
@ritik307 Yes exactly, but you can use `.js` extension, not `.jsx` :)

ritik307:
@magicmatatjahu I am currently facing so many errors while setting up the project. Can you kindly provide me with the steps to set up the project correctly? 

Following are the steps that I followed while setting up the project.

1. npm install -g @asyncapi/generator.
[...]

PR #448, opened 2023-09-11
feat!: switch to react, apiVersion 2 and newest react components

jonaslagoni:
**BREAKING CHANGE**
This PR adds a breaking change as we now this template depends on **generator `1.12.0` as the minimum**.

**Description**
This PR updates the HTML template to use react rendering and updates the template to use the new v2 `apiVersion` and updates the AsyncAPI react components.

**Related issue(s)**
Blocked by https://github.com/asyncapi/generator/pull/1017
Blocked by https://github.com/asyncapi/asyncapi-react/pull/766
Blocked by https://github.com/asyncapi/generator/pull/1039
Blocked by https://github.com/asyncapi/parser-js/pull/861
Fixes https://github.com/asyncapi/html-template/issues/430
Fixes https://github.com/asyncapi/html-template/issues/185


test/helpers/all.test.js
```diff
@@ -127,10 +137,13 @@ describe('Filters', () => {
...
-      const doc = new AsyncAPIDocument({ asyncapi: '2.0.0', info: { title: 'dummy spec for testing', version: '1.5.34' } });
-
-      const result = renderSpec(doc);
+    it.skip('should work', async () => {
+      const schema = new AsyncAPIDocumentV2({
+        asyncapi: '2.0.0',
+        info: { title: 'test', version: '0.0.0' },
+        channels: {},
+      });
+      const result = renderSpec(schema);
```
jonaslagoni, line 146: I unfortunately could not get the `renderSpec` part to work, which means that there will be no pre-rendering of the generated document (has to be loaded from the script).

Everything else works as expected.

jonaslagoni: @magicmatatjahu if you have a hunch or know why do let me know 😆

magicmatatjahu: Why it doesn't work? Any error?

jonaslagoni: It just renders empty string, no errors, nothing.

jonaslagoni: Also, why are we doing both the pre-rendering and then rendering it again in the script 🤔

magicmatatjahu: If you use a generator, the generator has a separate reference to AsyncAPIDocumentV2 and html-template has another - template and the generator is run in separate processes as well as in separate contexts so the dependencies are completely different. Pass a normal specification (as string or object) - properly parsed - not a model.

magicmatatjahu: or even stringified spec

magicmatatjahu: > Also, why are we doing both the pre-rendering and then rendering it again in the script 🤔

For SEO - first we render pure html without actions/handlers for clicks action and then only hydrate html - better SEO, because we have html and more faster than normal "rendering" od React.

jonaslagoni: > or even stringified spec

I tried both in the test (outside of generator) and it still renders `''`. Might need some longer debugging session to figure this out.

@magicmatatjahu do you see this as blocking the merge of this?

magicmatatjahu: > @magicmatatjahu do you see this as blocking the merge of this?

magicmatatjahu: > I tried both in the test (outside of generator) and it still renders ''. Might need some longer debugging session to figure this out.

But did you pass string or stringified spec? Also, maybe react component cannot recognize document inside https://github.com/asyncapi/asyncapi-react/blob/next/library/src/helpers/specification.ts#L15 this function

jonaslagoni: Thats the most likely cause yea...  

> But did you pass string or stringified spec

Stringified spec using `stringify(asyncapi)` and json with `asyncapi.json()`.

jonaslagoni: Tried to use the implementation with the generator and it seems to move it along.

But now the fun errors start 🤣 
```
Something went wrong:
Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.
    at resolveDispatcher (/Users/lagoni/Documents/github/asyncapi-react/library/node_modules/react/cjs/react.development.js:1465:13)
    at useState (/Users/lagoni/Documents/github/asyncapi-react/library/node_modules/react/cjs/react.development.js:1496:20)
    at AsyncApiLayout (/Users/lagoni/Documents/github/asyncapi-react/library/src/containers/AsyncApi/Layout.tsx:28:61)
    at processChild (/Users/lagoni/Documents/github/html-template/node_modules/react-dom/cjs/react-dom-server.node.development.js:3353:14)
    at resolve (/Users/lagoni/Documents/github/html-template/node_modules/react-dom/cjs/react-dom-server.node.development.js:3270:5)
    at ReactDOMServerRenderer.render (/Users/lagoni/Documents/github/html-template/node_modules/react-dom/cjs/react-dom-server.node.development.js:3753:22)
    at ReactDOMServerRenderer.read (/Users/lagoni/Documents/github/html-template/node_modules/react-dom/cjs/react-dom-server.node.development.js:3690:29)
    at Object.renderToString (/Users/lagoni/Documents/github/html-template/node_modules/react-dom/cjs/react-dom-server.node.development.js:4298:27)
    at renderSpec (/Users/lagoni/Documents/github/html-template/helpers/all.js:161:25)
    at Index (/Users/lagoni/Documents/github/html-template/components/index.js:10:24)
```

magicmatatjahu: So you:
- have two versions of react in single run
- component is broken and it tries use hook when it should be able not to.

You have to debug.

jonaslagoni: This might be because I am currently using symlink for the react components 🤔

jonaslagoni: Yep, works like a charm, just the test I need to adapt.

Thanks @magicmatatjahu 🙇

package.json
```diff
@@ -20,8 +20,8 @@
   "homepage": "https://github.com/asyncapi/html-template#readme",
   "scripts": {
     "lint": "echo \"No linter specified yet\"",
-    "develop": "npm run generate:assets && ag https://raw.githubusercontent.com/asyncapi/generator/v1.1.5/test/docs/dummy.yml ./ -o test/output --force-write --watch-template",
-    "//develop:install": "Force ag to install your local html-template source code",
+    "test:generator": "node ./node_modules/@asyncapi/generator/cli.js https://raw.githubusercontent.com/asyncapi/generator/v1.1.5/test/docs/dummy.yml ./ -o test/output --force-write",
```
derberg, line 23: we should use `@asyncapi/cli` and generator cli will be deprecated, removed and is no longer even promoted in docs

also since cli will be in devDependencies, afaik you should be able to run `asyncapi` without pointing to `node-modules`

derberg: yeah, I know you added generator, as cli did not support it yet, but yeah, now it does :)

package.json
```diff
@@ -93,14 +97,33 @@
         "required": false
       }
     },
-    "generator": ">=1.9.12 <2.0.0",
```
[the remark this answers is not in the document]
derberg: yes it does, which is not a problem

package.json
derberg: also, since you added 2 test files, v2 and v3 - can we use them here instead of remote one? both?

template/index.html.js
```diff
@@ -0,0 +1,10 @@
+import { File } from '@asyncapi/generator-react-sdk';
+import { Index } from '../components/index';
+
+export default function({ asyncapi, params = {} }) {
+  return (
+    <File name={params.outFilename || 'index.html'}>
```
derberg, line 6: > params.outFilename || 'index.html'

another lovely example about react power over nunjucks

derberg (changes requested):
left few comments,

please also add to description that this PR resolves https://github.com/asyncapi/html-template/issues/185 too

package.json
jonaslagoni: I can revert this change sure 🙂

package.json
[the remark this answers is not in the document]
jonaslagoni: Alright, switched to `feat!`.

fmvilas (dismissed):
LGTM 👍

jonaslagoni:
Need your review as well @derberg 🙂

derberg (commented):
left one small suggestion + also adjusted description a bit in regards to breaking change info, so it is clear for folks that will go into PR from release notes

.github/workflows/pr-testing-with-generator.yml
```diff
@@ -23,6 +23,6 @@ jobs:
           cache: 'npm'
           cache-dependency-path: '**/package-lock.json'
       - name: Install dependencies
-        run: NODE_ENV=production npm install
+        run: npm install
       - name: Run generation
-        run: npx -p @asyncapi/cli asyncapi generate fromTemplate https://raw.githubusercontent.com/asyncapi/generator/v1.1.5/test/docs/dummy.yml ./ -o test/output --force-write
+        run: npm run test:generator
```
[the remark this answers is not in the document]
derberg: please 🙏🏼 less custom workflows, I think we already see it is better 😄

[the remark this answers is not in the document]
jonaslagoni: Love it 👍

jonaslagoni:
@derberg you probably need to remove the required check 🙂

PR #574, opened 2024-03-21
fix: use latest JavaScript Parser version

KhudaDad414:
Resolves https://github.com/asyncapi/studio/issues/1031


derberg:
package-lock also should be updated

also please have a look at failing tests. This is probably why parser was not updated as tests were failing, so I closed a automated PR

KhudaDad414:
@derberg it should be good to go now. 👍

derberg (approved):

PR #575, opened 2024-03-23
feat: add scripts that on release will publish transpiled template to npm

### LINKED ISSUE #558: Add necessary scripts that on release will publish transpiled template to npm
Since we migrated this template to React render engine, the generation time, like in other react templates, got super long

Reason - every time generation is run, TS transpilation runs

This can be solved by publishing templates, like this one, to npm with transpilation files included, so basically no transpilation is needed in runtime - unless explicitly requested.

When you clone this repo, install and run `asyncapi generate fromTemplate ./test/spec/asyncapi_v3.yml ./ --output mydocs --force-write` from root, you will notice a new folder called `__transpiled` - this is the folder we need to make sure is included in package publishing.

This is how generator triggers transpilation -> https://github.com/asyncapi/generator/blob/master/lib/renderer/react.js#L19
Now we need to make sure we have proper script here in template that does the same. This new script needs to be called only by this script -> https://github.com/asyncapi/html-template/blob/master/package.json#L33C6-L33C20

We already have some custom scripts in this repo that run on release -> https://github.com/asyncapi/html-template/blob/master/scripts/copy-sources.js so you just plugin into existing infra.

**Note**
Completion of this task do not mean yet that time of generation will go down. I think that babel that we use in https://github.com/asyncapi/generator-react-sdk/blob/master/src/transpiler/transpiler.ts is already smart and do not regenerate if `__transpiled` dir is already there but yeah, lets see

utnim2:
**Description**
- Added the transpile scripts that creates `__transpiled ` folder in the root directory
- Modify the `prepublishOnly` scripts that ensures transpilation runs before publishing

**Related issue(s)**
Fixes #558

### DIFF
--- a/scripts/transpile.js
+++ b/scripts/transpile.js
@@ -1,11 +1,16 @@
-const { transpileFiles } = require('@asyncapi/generator-react-sdk');
-const path = require('path')
+const { transpileFiles } = require("@asyncapi/generator-react-sdk");
+const path = require("path");
 
 async function transpileTemplate() {
-  const templateContentDir = path.join(__dirname, '../template');
-  console.log("templateContentDir",templateContentDir)
-  const outputDir = path.join(__dirname, '../__transpiled');
-  await transpileFiles(templateContentDir, outputDir, {recursive: true})
+  try {
+    const templateContentDir = path.join(__dirname, "../template");
+    console.log("Template content directory:", templateContentDir);
+    const outputDir = path.join(__dirname, "../__transpiled");
+    console.log("Output directory for transpiled files:", outputDir);
+    await transpileFiles(templateContentDir, outputDir, { recursive: true });
+  } catch (error) {
+    console.log("Error during template transpilation:", err)
+  }
 }
 
-transpileTemplate().catch((err) => console.log('Error during template transpilation:', err));
+transpileTemplate();
--- a/scripts/transpile.js
+++ b/scripts/transpile.js
@@ -1,0 +1,15 @@
+const { transpileFiles } = require('@asyncapi/generator-react-sdk');
+const path = require('path')
+
+async function transpileTemplate() {
+  const templateContentDir = path.join(__dirname, '../template');
+  console.log("templateContentDir",templateContentDir)
+  const outputDir = path.join(__dirname, '../__transpiled');
+  console.log("outputDir",outputDir)
+
+  await transpileFiles(templateContentDir, outputDir, { recursive: true });
+}
+
+transpileTemplate().catch((err) => console.log('Error during template transpilation:', err));
+
+module.exports = { transpileTemplate }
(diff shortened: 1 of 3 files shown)


scripts/transpile.js
```diff
@@ -0,0 +1,19 @@
+const { transpileFiles } = require('@asyncapi/generator-react-sdk');
+const path = require('path')
+
+async function main(){
```
derberg, line 4: I don't think it's needed to wrap it like this in "main" function cause transpileTemplate is a wrapper already

scripts/transpile.js
```diff
@@ -0,0 +1,19 @@
...
+  async function transpileTemplate() {
+    const templateContentDir = path.join(__dirname, '../template');
+    console.log("templateContentDir",templateContentDir)
```
derberg, line 7: you can put nicer names for these logs

derberg (commented):
I don't think we actually need to modify copy sources script as we in the end publish entire package, not only template folder. Copy sources is needed only for specific thing.

So yeah, package.json script modification should be enough. Only one more change is needed - _trabspiled is in gitignore, so by default will not be pushed to npm,so you need to modify package.json so it accepts _trabspile folder

scripts/transpile.js
utnim2: 😅

scripts/transpile.js
utnim2: done 👍

scripts/transpile.js
```diff
@@ -0,0 +1,11 @@
+const { transpileFiles } = require('@asyncapi/generator-react-sdk');
+const path = require('path')
+
+async function transpileTemplate() {
+  const templateContentDir = path.join(__dirname, '../template');
+  console.log("templateContentDir",templateContentDir)
```
derberg, line 6: please add full sentence console log

scripts/transpile.js
```diff
@@ -0,0 +1,11 @@
...
+  const outputDir = path.join(__dirname, '../__transpiled');
```
derberg, line 7: console log this variable as well, but again, with full descriptive console log info

scripts/transpile.js
```diff
@@ -0,0 +1,11 @@
...
+  await transpileFiles(templateContentDir, outputDir, {recursive: true})
```
derberg, line 8: please put it in between try/catch instead of later catching errors on transpileTemplate

derberg (commented):
looks good, just some code improvements please

utnim2:
Done @derberg 👍

derberg (approved):
LGTM, lets merge and see if transpiled files end up in a package and if it makes any difference

aeworxet:
@asyncapi/bounty_team

Issue #256, opened 2021-09-08
remove puppeteer from dependencies


flamewow:
#### Reason/Context
Please try answering few of those questions
- Why we need this improvement?
**puppeteer is used only for PDF generation while being quite heavy (~200Mb)** 
- How will this change help?
**download time will be increased drastically**
- What is the motivation?
**PDF generation is not used very often and for users that are not using it at all 200Mb overhead is way too much**

#### Description
Please try answering few of those questions
- What changes have to be introduced?
**remove puppeteer from dependencies**
- Will this be a breaking change?
**yes**
- How could it be implemented/designed?
**change PDF generation function to accept puppeteer instance as a parameter. In that case, users that need PDF will have to install puppeteer separately and pass it as an argument**

magicmatatjahu:
@flamewow Hi! Thanks for the issue. Usually if you install a template once then the `puppeter` should be cached and reused in the next render. The problem may be the CI environment, where the `puppeter` will be installed with every job and I agree this can be frustrating.

I can't understand how do you want pass puppeteer instance as a parameter. Using Generator API it can be easy to do, but how do you want pass reference to the instance by the CLI? I only see solution that `puppeter` should be installed separately - before the template - globally and then we should use this global dep in the hook rather local instance.

derberg:
yeah, it is not really puppeteer that is heavy here but chromium and you can skip it as @magicmatatjahu mentioned.

> PDF generation is not used very often

unfortunately, we do not know that really

if using `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` is not something that satisfies you, the only good solution would be to completely remove PDF support from this template really and create a separate template like, `pdf-template`. Positive side effect would be better visibility of this option, so users would see they can generate markdown, HTML and PDF with 2 different templates. Thoughts? Of course we would need help here

flamewow:
@derberg creating a separate pdf-template package (will depend on html-template I would guess) sounds like a great option.
Ofc that's additional overhead for package management (and that's on you guys, to keep it under @asyncapi namespace in npm registry). Will be able to help with the code

derberg:
> Ofc that's additional overhead for package management (and that's on you guys, to keep it under @asyncapi namespace in npm registry)

this part is easy, we have releases and deps update inside asyncapi org fully automated.

> will depend on html-template I would guess

definitely. I haven't seen the code that generated pdf for some time but should be easy to port to new template and use `html-template` as dependency

> Will be able to help with the code

[...]

PR #668, opened 2024-08-28
chore: update puppeteer to 23.2.0

bakasmarius:
**Description**

- Puppeteer updated to the newest version (23.2.0)
- PDF generation tested with `-p pdf=true` flag
[index.new.v2.pdf](https://github.com/user-attachments/files/16782617/index.new.v2.pdf) - results of `test:generator:v2` with puppeteer@23.2.0
[index.new.v3.pdf](https://github.com/user-attachments/files/16782618/index.new.v3.pdf) - results of `test:generator:v3` with puppeteer@23.2.0
[index.old.v2.pdf](https://github.com/user-attachments/files/16782619/index.old.v2.pdf) - results of `test:generator:v2`  with puppeteer@14.4.1
[index.old.v3.pdf](https://github.com/user-attachments/files/16782620/index.old.v3.pdf) - results of `test:generator:v3` with puppeteer@14.4.1
The results do not match pixel-perfectly, but the differences are very minor (one-pixel size margin/padding here and there), so I don't think it matters in this case.

**Related issue(s)**
Requested during discussion in #256

### DIFF
--- a/hooks/99_generatePdf.js
+++ b/hooks/99_generatePdf.js
@@ -2,7 +2,7 @@ const puppeteer = require('puppeteer');
 const path = require('path');
 
 /**
- * Generates PDF if user pass `pdf` paramater.
+ * Generates PDF if user passes `pdf` parameter.
  */
 module.exports = {
   'generate:after': async ({ templateParams = {}, targetDir }) => {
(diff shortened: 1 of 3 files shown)


derberg:
thanks! changed to `fix` as with major releases of dependencies we do patch releases

PR #686, opened 2024-10-02
feat: render react app setup into dedicated js file

mwiesenbauer:
The HTML Template currently renders inline javascript with schema and react application setup code.
Inline javascript in combination with a content security policy (CSP) is discouraged.

Rendering schema and application setup code into a dedicated javascript file will resolve this issue.

`singleFile` option is still supported with this change


derberg:
lgtm, but please make sure we still support an option that all files can be also inlined in one file with `singleFile` property

derberg (commented):
Thanks a lot for all the work.

I also tested manually:
```
### default
asyncapi generate fromTemplate  ./test/docs/dummy.yml https://github.com/mwiesenbauer/html-template -o ./dupa --force-write --use-new-generator
```
and
```
### single file
asyncapi generate fromTemplate  ./test/docs/dummy.yml https://github.com/mwiesenbauer/html-template -o ./dupa --force-write --use-new-generator -p singleFile=true
```

I have only once concern - if this is not a breaking change.

in theory, there is no change in functionality, but imho output is part of template "functionality", especially in case of code/docs generators. This new html-template version will output by default one extra file - MUST HAVE file. This may affect pipelines if people configure them explicitly to react only on certain output files by their name, you know? like copy not entire directory output but just certain files, for various reasons - maybe security.

so I'm thinking if we should not do major here

any thoughts?

derberg (commented):
Ah, forgot you need to update tests. Notice they are failing. These are snapshot test. Problem is that in tests we test directly `Index` function but it's output is now different, so you will need to update the snapshot test to reflect this change, and push to that PR for review. When running tests locally, just add manually `--updateSnapshot` flag to [script where jest is invoked](https://github.com/asyncapi/html-template/blob/master/package.json#L23) (you can also add new script for it and push to this PR. Also you need to extend the snapshot test by also adding a code verification of the new `App` component rendering.

mwiesenbauer:
@derberg Sorry for the delay, done.

derberg (approved):

derberg:
thanks a lot for the PR and patience with getting it merged.

I got no response from community on my major/minor concerns, so this time we go with minor 🤷🏼 

/rtm

PR #689, opened 2024-10-18
docs: update README.md to master React Component Docs

maxbin123:
Let's update link in readme to actual version of React Component docs, it has more options now


derberg (approved):
Thanks!

PR #708, opened 2025-02-13
feat: `@asyncapi/react-componet` upgraded to v2 together with `react`

### LINKED ISSUE #707: Upgrade @asyncapi/react-component dependency major version from 1 to 2
### Why do we need this improvement?

`@asyncapi/html-template` seems to have fallen behind in development and has not been taking advantage of fixes in `@asyncapi/react-component` since April 2024. The react-component dependency is using the old version ^1.2.13, which was released in January 2024, and in April 2024, the major version 2.0.0 of react-component was released, which upgraded the component to React v18.

### How will this change help?

An example: react-component v2.2.2 fixed an issue where hiding a section would not hide it from sidebar. Since html-template is still using the major version 1, if you generate an API document using CLI with { "show": { "schemas": false } } in the config file, the schemas section will not be visible in the document, but will still appear in the sidebar. Issues like these would be fixed by bumping the major version to 2.

### Screenshots

_No response_

### How could it be implemented/designed?

Upgrading the `@asyncapi/react-component` dependency and ensuring that this change does not break html-template. The upgrade to React v18 in react-component is listed as a breaking change, which is why it's a major update.

### 🚧 Breaking changes

Yes

- [X] I checked and didn't find a similar issue

- [X] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue?

No, someone else can work on it.

neoandmatrix:
**Description**

This PR introduces the changes to upgrade the `@asyncapi/react-component` to it's latest version `2.5.1`. The extra added dependency `react` is to allow installation of newer `react-dom` as required by `@asyncapi/react-component`.

The tests are updated as well and are all passing.

image

**Related issue(s)**
Resolves #707 

@derberg please review the PR and provide feedback.

Thanks.


derberg (approved):
lgtm, thanks a lot

neoandmatrix:
> @neoandmatrix please look into failing tests

Sure, will check and update.

Thanks.

neoandmatrix:
Hey @derberg was working on the failing tests and after some debugging found the cause to be this method 
image at `helpers/all.js` in `renderSpec` function.  The corresponding error `Generator Error: window is not defined ` might be due to changes in newer version of react that might require browser functions or some `dom functions`. I was initially suspicious that this might be the as we are cause as we are using `react-dom/server` but just to confirm i tried using `jsdom` library to simulate the required browser functions as 
image

and the generator worked as expected and correct output files were being generated. But this caused a new problem that after generating the template the generator is stuck and fails to exit only only exits upon using `ctrl+c` to manually exit it.

Can you please provide some feedback on if my approach is correct and if so then what might be probable cause of this being not exiting on itself. If some docs or material should be that i can refer please provide that will be much helpful for me.

Thanks.

derberg:
@neoandmatrix you need to push your changes to PR so I can have a closer look

derberg:
I tried your changes on local, using `npm run test:generator:v3` and your solution works, and cli terminates without issues.

also looks like tests are passing here too

neoandmatrix:
> I tried your changes on local, using `npm run test:generator:v3` and your solution works, and cli terminates without issues.
> 
> also looks like tests are passing here too

Oh that's good then, looks like some older cache might be causing issue on my system then.

derberg (approved):

derberg:
@all-contributors please add @neoandmatrix for code

Issue #720, opened 2025-04-09
[BUG] Parallel execution command asyncapi generate got error for the SingleFile set True


AayushSaini101:
### Describe the bug.

Reference to this issue: https://github.com/asyncapi/cli/issues/814 Parallel execution is causing errors for the case of the HTML template when the singleFile=true 

### Expected behavior

Fix Attempt: https://github.com/asyncapi/html-template/blob/7188fbf2c4ab1f061f62fd6fc983e2c737f59201/hooks/01_removeNotRelevantParts.js#L25 Add a check here to validate whether the files have been removed or not to prevent not found errors during the generation 

### Screenshots

<img width="1885" alt="Image" src="https://github.com/user-attachments/assets/f0522dde-7088-4671-b633-f4159cc3dde7" />

### How to Reproduce

Steps to Produce: https://github.com/asyncapi/cli/issues/814#issuecomment-1737706238

### 👀 Have you checked for similar open issues?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [x] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue ?

Yes I am willing to submit a PR!

PR #721, opened 2025-04-09
fix: add a check to fix the generation failed in the parallel execution

AayushSaini101:
Resolves: #720


derberg (commented):
did you test it? how this can work without requiring `fs` first

AayushSaini101:
> did you test it? how this can work without requiring `fs` first

@derberg my bad added thanks

PR #735, opened 2025-07-15
fix: coerce singleFile strings->bools so non-singleFile won't bundle deps

### LINKED ISSUE #734: [BUG] Inconsistent singleFile behaviour
### Describe the bug.

Currently, using the HTML-Template to generate docs, and using the `singleFile: 'false'` option behaves very weirdly. If it's either `'true'` or `true` then it's fine, but if it's `'false'` or `false` it doesn't behave as you would expect in either case.

The docs for @asyncapi/html-template state this value should be a string.

If it's a string of `'false'` (which is a truthy value - likely partially the source of the issue), it incorrectly generates the `index.html` as the singleFile format - the same as if you were to use a truthy value. However, it ALSO generates the docs as external dependencies, as you would expect it to when the value is set to `false`.

If it's a boolean of `false` then it correctly generates the file without the dependencies inside of it, like you would expect. However, it does not generate the external dependencies which it needs to work.

**Tested using versions:**
Node v22.2.0
@asyncapi/generator: 2.7.0
@asyncapi/parser: 3.4.0
@asyncapi/html-template on 2025-07-11 (so probably pulling v3.3.0)

#### Stretch goal:

It would also be a nice value add to me if there were additional options to:
1)  _intentionally_ generate the minimal documentation without the external dependencies, for the use case where I'm generate a bunch of different documentation and wish to point it all to the same self-hosted JS/CSS dependencies. This means I can save space and time when generating docs.
2) generate only the dependencies, which I can run when I build and deploy the server I host those dependencies on.

I understand my use case may not be worth specifically catering to. I can cope with that, it's just a nice-to-have.

### Expected behavior

Both `false` and `'false'` behave the same, and generate the minimal documentation which relies on external deps, and also generate those deps.

### Screenshots

Image

Image

### How to Reproduce

https://github.com/michael-ball-ctct/async-api-generator-singlefile-bug-demo

1. Clone minimal repoduction and navigate to its directory.
2. `npm install`.
3. `npm run start`.
4. Read test output/inspect files generated in `./docs`

- Operating System (OS): Ubuntu 22.04.5 LTS
- Browser: N/A
- Browser Version: N/A

Yes I am willing to submit a PR!

**NB:** In theory, I'm happy to contribute a PR to help push this along. In practice, I'm not really even sure where to start with this as I don't have a deep understanding of the internals of AsyncAPI, and I'm unsure how much time I'll be able to dedicate to this.

michael-ball-ctct:
**Description**

Currently html-template parses the string `'false'` as `true` because it is a non-empty string. This leads to the html file that is generated being created with its dependencies built in, when it should not.

By coercing this value to a boolean, this change ensures that only strings that match `'true'` or true boolean values will match the singleFile behaviour, and otherwise stick to the default non-singleFile behaviour. Other strings (including `'false'`) will generate non-singleFile documentation.

This behaviour satisfies the TypeScript contract which states that the value of the `singleFile` optional param must be a string. It does not solve the behaviour that if AsyncAPI/generator is given a singleFile param of a `false` boolean it will generate a non-singleFile formatted `index.html` while failing to also generate the dependency files non-singleFiles refer to. This is because A) fixing this behaviour appears to be outside of the html-template repo and B) that's outside of the aforementioned expected types supported, as per the TypeScript definition.

Using the [test repo](https://github.com/michael-ball-ctct/async-api-generator-singlefile-bug-demo) I created for #734 and pointing the template towards this repo will show a fixed case for `singleFileFalseString` (but not `singleFileFalseBoolean`, as discussed above).

**Related documentation**
None, other than the code comment explaining the coercion - which can be removed/amended if not fitting with repo style. This change is ensuring behaviour matches existing TypeScript + written documentation.

**Related issue(s)**

Fixes #734 

**Test results**
```
 PASS  test/helpers/all.test.js
 PASS  test/components/index.test.js
------------|---------|----------|---------|---------|-----------------------------
File        | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s           
------------|---------|----------|---------|---------|-----------------------------
All files   |      70 |    66.66 |     100 |   70.88 |                             
 components |   76.47 |       50 |     100 |   76.47 |                             
  index.js  |   76.47 |       50 |     100 |   76.47 | 15,22,29,33                 
 helpers    |   68.25 |    73.33 |     100 |   69.35 |                             
  all.js    |   68.25 |    73.33 |     100 |   69.35 | 18,22,45-50,103-117,163-166 
------------|---------|----------|---------|---------|-----------------------------

Test Suites: 2 passed, 2 total
Tests:       13 passed, 13 total
Snapshots:   4 passed, 4 total
Time:        2.707 s
Ran all test suites.
```

### DIFF
--- a/components/index.js
+++ b/components/index.js
@@ -9,14 +9,16 @@ export function Index({ asyncapi, params = {} }) {
   const favicon = generateBase64Favicon(params);
   const renderedSpec = renderSpec(asyncapi, params);
   let asyncapiScript = `<script src="js/asyncapi-ui.min.js" type="application/javascript"></script>`;
-  if(params?.singleFile) {
+  // coerce singleFile param to bool, or "false" string will be true
+  const singleFile = (params?.singleFile === true  || params?.singleFile == 'true');
+  if(singleFile) {
     asyncapiScript = `<script type="text/javascript">
     ${includeFile('template/js/asyncapi-ui.min.js')}
     </script>`;
   }
   let styling = `<link href="css/global.min.css" rel="stylesheet">
       <link href="css/asyncapi.min.css" rel="stylesheet">`;
-  if(params?.singleFile) {
+  if(singleFile) {
     styling = `<style type="text/css">
       ${includeFile("template/css/global.min.css")}
       ${includeFile("template/css/asyncapi.min.css")}
@@ -27,7 +29,7 @@ export function Index({ asyncapi, params = {} }) {
     basehref = `<base href="${params.baseHref}">`;
   }
   let appJs = `<script type="application/javascript" src="js/app.js"></script>`;
-  if(params?.singleFile) {
+  if(singleFile) {
     appJs = `<script>${App({asyncapi, params})}</script>`;
   }
   return (`<!DOCTYPE html>


derberg (approved):
lgtm, thanks a lot!

PR #754, opened 2025-11-25
fix: missing css and js files while singleFile=false

### LINKED ISSUE #744: [BUG] Missing CSS/JS when generating without `singleFile`
### Describe the bug.

When generating docs with `@asyncapi/html-template` version `3.3.0+`, the generated `index.html` references CSS and JS files, but they are not created in the output folder when singleFile is not used.

With `3.2.0`, both `singleFile=true` and `singleFile=false` modes worked correctly. Starting with `3.3.0`, only `singleFile=true` works.

>[!TIP]
> **I suspect the issue is with the new conditional generation added.**

### Expected behavior

- If singleFile=true: generate one standalone HTML file with bundled CSS/JS (works fine in 3.3.1 also).
- If singleFile=false (or not passed): generate index.html + referenced css/ and js/ assets (broken in 3.3.0+).

### Screenshots

Image

### How to Reproduce

1. Run:  `./bin/run generate fromTemplate asyncapi.yaml @asyncapi/html-template -o ../docs --use-new-generator --force-write`

2. Example spec (simplified):

```yaml
asyncapi: 3.0.0
info:
  title: Account Service
  version: 1.0.0
channels:
  userSignedup:
    address: user/signedup
    messages:
      UserSignedUp:
        $ref: '#/components/messages/UserSignedUp'
components:
  messages:
    UserSignedUp:
      payload:
        type: object
        properties:
          email:
            type: string
```

3. Open the generated index.html in the browser.

4. Notice that it references CSS/JS files, but none are generated.

- Operating System (OS):
- Browser:
- Browser Version:

None

batchu5:
CSS and JS files are missing because in conditionalGeneration the value of the const should be false not "false" 

in generator also there is bug in one line 

 const defaultValues = Object.keys(parameters || {}).filter(key => parameters[key].default); 
 should be changed to 
 const defaultValues = Object.keys(parameters || {}).filter(key => parameters[key].default !== undefined);
 
 because just .default considers the false value as false and doesn't filter out!!
fixes #744

### DIFF
--- a/.github/workflows/automerge-for-humans-add-ready-to-merge-or-do-not-merge-label.yml
+++ b/.github/workflows/automerge-for-humans-add-ready-to-merge-or-do-not-merge-label.yml
@@ -27,6 +27,8 @@ jobs:
     steps:
       - name: Add ready-to-merge label
         uses: actions/github-script@v7
+        env:
+          GITHUB_ACTOR: ${{ github.actor }}
         with:
           github-token: ${{ secrets.GH_TOKEN }}
           script: |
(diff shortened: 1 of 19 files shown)


lightning-sagar:
@batchu5 i believe the issue goal, is when u run `-p singleFile=false` the js/css file is not being generated. i dont think it's generating the css/js

PR #757, opened 2025-11-27
fix: default value of the singleFile is changed

### LINKED ISSUE #744: [BUG] Missing CSS/JS when generating without `singleFile`

batchu5:
**Description**
In generator if we give the params as -p single=false eventually, it is converting into single = 'false' , that is changed into singleFile='false', so changing the default value from false to 'false' in package.json results in string and css and js files are generated

Fixes #744


batchu5:
okay, can you explain me where the generator is not handling the default false, 
I tried to debug and found the error, same as where you raised the PR (in loader.js), you mentioned it is only handling singleFile="false".

and also i think params should be false not "false". there is no need to handle for both cases

